PR #2, opened 2017-12-19
[MIG] base_external_system: Upgrade to v11

lasley:
Perform standard v11 migration. Includes OCA/server-tools#1033 as 4db2a3e. Moved from OCA/server-tools#1097


pedrobaeza:
The contributor is no longer active, but this migration is pretty straight-forward, so I think we can merge it.

PR #3, opened 2018-02-21
[IMP] Use default user to define default roles to create for a new user

benwillig:


sebalix (approved):
Port of https://github.com/OCA/server-tools/pull/1024, :+1:

andreparames (approved):

PR #10, opened 2018-07-12
[11.0][FIX] suspend security incorrect uid in env

katyukha:
When `sudo` is used, then Odoo converts `uid` to int, thus later it is not recognized as `BaseSuspendSecurityUid` instance

So, to avoid this conversion, new environment created (similary to that is happening in `sudo` call)

### DIFF
--- a/base_suspend_security/__manifest__.py
+++ b/base_suspend_security/__manifest__.py
@@ -18,7 +18,7 @@
 ##############################################################################
 {
     "name": "Suspend security",
-    "version": "11.0.1.0.0",
+    "version": "11.0.1.0.1",
     "author": "Therp BV, brain-tec AG, Odoo Community Association (OCA)",
     "license": "AGPL-3",
     "category": "Hidden/Dependency",
--- a/base_suspend_security/models/base.py
+++ b/base_suspend_security/models/base.py
@@ -12,4 +12,8 @@ class Base(models.AbstractModel):
 
     @api.model
     def suspend_security(self):
-        return self.sudo(user=BaseSuspendSecurityUid(self.env.uid))
+        return self.with_env(
+            api.Environment(
+                self.env.cr,
+                BaseSuspendSecurityUid(self.env.uid),
+                self.env.context))


hbrunn:
are you sure you're not fixing a symptom? `BaseSuspendSecurityUid` [overrides casting to int](https://github.com/crnd-inc/server-backend/blob/6c48ea9867ae0b0365ae7d079dbca45a51b9b5bc/base_suspend_security/base_suspend_security.py#L23) for exactly that reason.

katyukha:
Just short test in IPython:

```ipython
In [1]: class BaseSuspendSecurityUid(int):
   ...:     def __int__(self):
   ...:         return self
   ...:     def __eq__(self, other):
   ...:         if isinstance(other, int):
   ...:             return False
   ...:         return super(BaseSuspendSecurityUid, self).__int__() == other
   ...:     def __hash__(self):
   ...:         return super(BaseSuspendSecurityUid, self).__hash__()
   ...:     def __iter__(self):
   ...:         yield super(BaseSuspendSecurityUid, self).__int__()
   ...:         

In [2]: c = BaseSuspendSecurityUid(5)

In [3]: type(c)
Out[3]: __main__.BaseSuspendSecurityUid

In [4]: type(int(c))
./projects/venv/bin/ipython:1: DeprecationWarning: __int__ returned non-int (type BaseSuspendSecurityUid).  The ability to return an instance of a strict subclass of int is deprecated, and may be removed in a future version of Python.
  #!/home/katyukha/projects/venv/bin/python3
Out[4]: int
```

Python and IPython versions:

```
Python 3.6.5 (default, Apr  1 2018, 05:46:30)
IPython 6.3.1
```

Odoo converts `uid` to `int` in `api.Environment.__call__` method.
So, in this commit new environment created directly, to avoid calling environment.

hbrunn:
thanks for the elaboration. I just checked this with python 3.6.6, and there I neither get the warning nor the behavior you showed. Weird stuff.

But I was merely curious, your patch seems functionally equivalent to me, so if it benefits you and harms nobody else, we can just as well do it the way you propose I think

hbrunn (approved):

katyukha:
Warning was shown when i run example from IPython shell, but when i run it from Python shell, no warning is shown, but type of `int(BaseSuspendSecurity(5))` is `int`.

hbrunn:
that's the weird part, on my python 3.6.6, it's BaseSuspendSecurity

hbrunn:
but no worries, now we have a version that works for both cases, thanks

katyukha:
thanks for review)

pedrobaeza:
Please change module version number according https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#version-numbers

katyukha:
Module version changed

BT-fgarbely:
Hello @pedrobaeza 

What is the state of this pull request?
Can you merge it?

@katyukha 
Thanks for the fix. It works now also on my side (tested in python 3.5.5 and python 3.6.5)

pedrobaeza:
Travis and runbot are still red.

katyukha:
@pedrobaeza @BT-fgarbely rebased, travis and runbot now green.

pedrobaeza (approved):

PR #25, opened 2019-01-02
[base_user_role] improve performance by only adding/removing from a user the groups that have changed with respect to the role.

JordiBForgeFlow:


JordiBForgeFlow:
@sebalix @alexis-via We badly need this PR for performance reasons. On an environment with various hundreds of users and many roles, using an update of groups with [(6,0,..)] was too bad.

base_user_role/models/user.py
```diff
@@ -55,6 +55,11 @@ def set_groups_from_roles(self, force=False):
         If no role is defined on the user, its groups are let untouched unless
         the `force` parameter is `True`.
         """
+        role_groups = {}
```
sebalix, line 58: Can you put a comment describing why you need to do that? For further reviews/migrations!

sebalix:
@jbeficent LG, have you some "benchmarks" regarding this?

JordiBForgeFlow:
@sebalix Yes. We tested 650 users and 30 roles. The biggest problem was a role with more than 300 users.

It was taking about 3 minutes to complete, and after this change it takes about 20 secs.

base_user_role/models/user.py
JordiBForgeFlow: We obtain all the groups associated to each role first, so that it is faster to compare later with each user's groups.

JordiBForgeFlow: I am adding this as a comment

JordiBForgeFlow:
@sebalix Plus, most of the times when the cron is executed there won't be any change in groups, so it's going to take just a few seconds to compare the roles groups with user's groups and decide that there's nothing to update.

The real performance killer was https://github.com/odoo/odoo/blob/12.0/odoo/addons/base/models/res_users.py#L853 when you decided to update all groups for all users.

etobella (approved):
LGTM :+1:

sebalix (approved):

PR #26, opened 2019-01-13
[ADD] base_dav

hbrunn:
This module adds WebDAV support to Odoo, specifically CalDAV and CardDAV.

You can configure arbitrary objects as a calendar or an address book, thus make arbitrary information accessible in external systems or your mobile.

Configuration
=============

* go to `Settings / WebDAV Collections` and create or edit your collections. There, you'll also see the URL to point your clients too.

Note that you need to configure a dbfilter if you use multiple databases.

Known issues / Roadmap
======================

* better UX for configuring collections
* support writing
* support address books
* support todo lists and journals
* support configuring default field mappings per model
* support plain WebDAV collections to make some model's domain accessible as folders, and the records' attachments as files (r/w)
* support configuring lists of calendars so that you can have a calendar for every project and appointments are tasks, or a calendar for every sales team and appointments are sale orders. Lots of possibilities

Backporting this to <=v10 will be tricky because radicale only supports python3. Probably it will be quite a hassle to backport the relevant code, so it might be more sensible to just backport the configuration part, and implement the rest as radicale auth/storage plugin that talks to Odoo via odoorpc. It should be possible to recycle most of the code from this addon, which actually implements those plugins, but then within Odoo.


pedrobaeza:
Shouldn't this be `base_webdav`, as both protocols start from this base protocol? I was doubting about putting this on `OCA/calendar`, but `CardDAV` part is not pure calendar...

hbrunn:
well, it supports some dav-protocols, so I figured this name will do. But I can rename it, no problem.

This still needs quite an amount of work to be really useful (it's readonly for now, and needs work to support things like pictures or complex names), so better to rename now than later

hbrunn:
I just configured a calendar on runbot which is accessible at http://3353937-26-b656c9.runbot3.odoo-community.org/.dav/admin/1 for as long as this instance lives:

image

Then after configuring this in horde as caldav client, it looks like this:

image

Those are some of the events in Odoo's calendar, yay! A lot of stuff still goes wrong, but that's a good start I think. And nobody forces us to use `calendar.event` as model, you can configure whatever you want there. But you need to configure at least fields `dtstart` and `summary` to see something sensible on the receiving end.

ping @StefanRijnhart @thomaspaulb

fkantelberg:
@hbrunn 
Thanks for the work.

Can you have a look on [this](https://github.com/hbrunn/server-backend/pull/1)? It should add the first upload/delete functionality. During the testing I noticed an interesting behaviour. If you setup a _all day event_ this will result in caldav with end date = start date + 1 day and odoo will interpret it as 2 days which shows the limitation of the simple mapping.

I kinda think that we could provide text fields for import/export with python code. We can just interpret them with safe_eval and they should convert from dict to/from record while the framework will convert the dict to/from vobject.

hbrunn:
@fkantelberg thanks, reviewed it including a guess about the two days problem.

The python expression for generating field values seems sensible to me, but note that in this case we'll also need a code snippet to parse a value again for the write case. And another column to determine the type in case it's not the code's responsibility to return exactly what is needed. We can expose some helper functions in the evaluation context to simplify this. In any case, the mapping then should have a type column where the user can switch between 1:1 field mappings and code. I think that's better to contribute after version 1 is merged.

Last thing we need to get this mergeable are tests. If you want to do the honors, you can yank the context manager from https://github.com/OCA/server-auth/pull/131/files#diff-c2041f2f2bc5a25a8ce303fb25923980R21 and adapt it to set the headers for basic authentication and whatever else caldav needs. That's better than just calling the server via HTTP, because those requests are handled in a different transaction, so you won't be able to assert anything about stuff happening. It should be possible to mock https://github.com/python-caldav/caldav/blob/master/caldav/davclient.py#L243 and have the `caldav` library do the grunt work of creating the correct request data.

hbrunn:
or if that looks too complicated, tests just using the `caldav` library and asserting things about a write/read roundtrip should be good enough for starters. But don't depend on `calendar` for the test data, better add a caldav configuration for some object available in the `base` module.

fkantelberg:
@hbrunn  I was thinking for the python code conversion to do this on the dav.collection level and not within the field mapping lines because you might want to use different fields for something like summary. I'm currently testing it and it looks promising.

I might be able to check out the tests next week. I would create a new PR for this towards your branch.

hbrunn:
I wouldn't advise making this part of the `dav.collection` model. If you add fields `mapping_type`, `code_to_vobject` and `code_from_vobject` to the mapping model, and then adapt the `_{to,from}_vobject` functions to call this depending on the `mapping_type`, you'll be done. The `code_to_vobject` snippet can be passed just the `vobj`, and `code_from_vobject` the values dict (`result`). Expose the collection in both as `self`, then this should feel quite natural to use

hbrunn:
did you mean doing basically what I describe but at the end/before to/from vobject? So that we just add code_{to,from}_vobject which is run then? That's probably even simpler, and we can mix this with the field mapping approach. That's good I think.

fkantelberg:
Basically my idea is to split the functions `_to_vobject` into parts. The first part will be the field mapping or the execution of the code snippet stored at dav.collection. Both ways will result in a dictionary structure which will be converted to vobject. For the `_from_vobject` the parts are just swapped. I'll try to push the poc today.

hbrunn:
is the or exclusive? I think it will be very useful to allow both simultaneously and merge the resulting dicts

fkantelberg:
Yeah I meant it as exclusive because merging is kinda difficult. Maybe we can have both and just pass the mapped value into the python code aswell.

Currently I have [this](https://github.com/hbrunn/server-backend/pull/2) which goes with your suggestion.

hbrunn:
superseded by https://github.com/OCA/server-backend/pull/61

Issue #36, opened 2019-04-13
base_suspend_security: BaseSuspendSecurityUid.__iter__ method. Do we need it?


katyukha:
Hello!

Suspend Security wrapper currently has method `__iter__`, and that's why following [expression](https://github.com/odoo/odoo/blob/11.0/odoo/osv/expression.py#L957) evaluates to True: `elif isinstance(right, collections.Iterable):`. But later, following [code](https://github.com/odoo/odoo/blob/11.0/odoo/osv/expression.py#L443) tries to split that iterable by chunks, and raises error: `TypeError: object of type 'BaseSuspendSecurityUid' has no len()`.

So, do we need to keep `__iter__` method and also add `__len__` (and possibly some other methods)? Or we do not need this wrapper to look like iterable, and we could safelly remove this `__iter__` method?

katyukha:
Closing, there is PR #37 that fixes it.

PR #37, opened 2019-04-13
[FIX] base_suspend_security when uid wrapper is used in One2many domain

katyukha:
Fix related to issue #36 

Suspend Security wrapper currently has method `__iter__`, and that's why following [expression](https://github.com/odoo/odoo/blob/11.0/odoo/osv/expression.py#L957)  evaluates to `True`:

```python
elif isinstance(right, collections.Iterable):
```
But later, following [code](https://github.com/odoo/odoo/blob/11.0/odoo/osv/expression.py#L443) tries to split that iterable by chunks, and raises error: 

```TypeError: object of type 'BaseSuspendSecurityUid' has no len().```

This PR fixes bug described above.

### DIFF
--- a/base_suspend_security/__manifest__.py
+++ b/base_suspend_security/__manifest__.py
@@ -18,7 +18,7 @@
 ##############################################################################
 {
     "name": "Suspend security",
-    "version": "11.0.1.0.1",
+    "version": "11.0.1.0.2",
     "author": "Therp BV, brain-tec AG, Odoo Community Association (OCA)",
     "license": "AGPL-3",
     "category": "Hidden/Dependency",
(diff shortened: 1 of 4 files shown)


katyukha:
@pedrobaeza Hello, can you review this PR?

pedrobaeza:
Sorry, but I don't know this module enough for reviewing it.

hbrunn (changes requested):
don't remove `__iter__`, we really want to hit this branch in `_normalize_ids`.

So you could just say `('...id', '=', [self.env.uid])` in your domain to work around the issue. But if you want your case to work, better implement `__len__` and `__item__`

PR #39, opened 2019-05-30
[IMP] Adding a new tab for user comments in roles form.

HaraldPanten:
With this new tab, it's possible to describe each role in order to let other users know all details.


HaraldPanten:
@pedrobaeza I'm trying to preserve commit history but I can't. 

It's the same branch and I tried:

https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-11.0

But I don't know why the format-path is not collecting the commits.. How can I solve this?

Thanks in advance.

pedrobaeza (approved):
No problem in commit history as this is a simple feature and not sure if this is already in other branch.

ValentinVinagre (approved):
LGTM

PR #43, opened 2019-08-22
[ADD] base_global_discount: New Module 11.0

chienandalu:
Base module to have global discounts applied to either sales or purchases. It
doesn't do much for itself, so account_global_discount or
purchase_global_discount should be installed to benefit from it.

cc @Tecnativa


base_global_discount/__manifest__.py
```diff
@@ -0,0 +1,22 @@
+# Copyright 2019 Tecnativa S.L. - David Vidal
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+{
+    'name': 'Base Global Discount',
+    'version': '11.0.1.0.0',
+    'category': 'Point of Sale',
```
pedrobaeza: Wrong category

base_global_discount/models/global_discount.py
```diff
@@ -0,0 +1,62 @@
...
+    _order = "sequence, id desc"
+
+    sequence = fields.Integer(
+        help='Gives the order to apply discounts',
+    )
+    name = fields.Char(
+        string='Discount Name',
+        required=True,
+    )
+    color = fields.Integer('Color Index', default=10)
```
pedrobaeza: Why this?

base_global_discount/models/global_discount.py
```diff
@@ -0,0 +1,62 @@
...
+        selection=[
+            ('amount', 'Fixed amount'),
+            ('percentage', 'Percentage'),
+        ],
+        default='percentage',
+        required='True',
+        string='Discount Type',
+    )
+    discount = fields.Float(
+        digits=dp.get_precision('Product Price'),
```
pedrobaeza: This is defined in `product` module, so you must depend on it.

base_global_discount/models/global_discount.py
```diff
@@ -0,0 +1,62 @@
...
+        default=0.0,
+    )
+    discount_scope = fields.Selection(
+        selection=[
+            ('sale', 'Sale'),
```
pedrobaeza: s/Sale/Sales

base_global_discount/models/res_partner.py
```diff
@@ -0,0 +1,22 @@
+# Copyright 2019 Tecnativa - David Vidal
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from odoo import fields, models
+
+
+class ResPartner(models.Model):
+    _inherit = 'res.partner'
+
+    customer_global_discount_ids = fields.Many2many(
```
pedrobaeza, line 9: For a good visualization, add a `name_get` on `global.discount` with "Name (amount%)" or "Name (-amount)" depending on the type of discount.

base_global_discount/readme/USAGE.rst
```diff
@@ -0,0 +1,14 @@
+To use this module, you need to:
+
+#. Go to *Settings > Parameters > Global Discounts*
```
pedrobaeza: This first paragraph should go in CONFIGURE.rst

base_global_discount/models/global_discount.py
```diff
@@ -0,0 +1,62 @@
...
+            ('purchase', 'Purchase'),
+        ],
+        required='True',
+        string='Discount Scope',
+    )
+    company_id = fields.Many2one(
+        comodel_name='res.company',
+        string='Company',
```
pedrobaeza, line 35: Add default to current user company

base_global_discount/static/description/icon.svg
```diff
@@ -0,0 +1,79 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
pedrobaeza: Don't add SVG if it's the standard one

pedrobaeza (changes requested):
Add Spanish translation at least

base_global_discount/models/global_discount.py
chienandalu: In a first moment I thought it could be nice and cheap to have it in the m2m tags but in a second though it looks lame in this case. Removing it

base_global_discount/models/global_discount.py
chienandalu: Should not they be globally available by default?

base_global_discount/models/global_discount.py
chienandalu: done

base_global_discount/models/res_partner.py
chienandalu: done

base_global_discount/models/global_discount.py
pedrobaeza: Not really, and this is not the default behavior in other places. Imagine this scenario: you have a multi-company DB, and you are a user with permissions for only one company and you are not even aware of that field (as it's hidden by default with the multi-company group). You create one record, thinking that is only for your company, and it appears in other companies! So it should be the contrary: belonging to one company by default, and having to be changed to empty for doing it global.

base_global_discount/models/global_discount.py
```diff
@@ -0,0 +1,76 @@
...
+    )
+
+    def name_get(self):
+        types = {
+            'percentage': '({:.2f}%)',
+            'amount': '(-{:.2f})',
+        }
+        result = []
+        for one in self:
+            type = ''
```
pedrobaeza: Don't use this Python reserved word

base_global_discount/views/global_discount_views.xml
```diff
@@ -0,0 +1,49 @@
...
+                <field name="company_id"/>
+            </tree>
+        </field>
+    </record>
+
+    <record id="global_discount_view_form" model="ir.ui.view">
+        <field name="model">global.discount</field>
+        <field name="arch" type="xml">
+            <form string="Global Discounts">
+                <group col="4">
```
pedrobaeza: Add `<sheet>`

base_global_discount/models/global_discount.py
chienandalu: ok then

chienandalu:
All changes done

base_global_discount/models/global_discount.py
```diff
@@ -0,0 +1,58 @@
...
+        comodel_name='res.company',
+        string='Company',
+        default=lambda self: self.env.user.company_id,
+    )
+
+    def name_get(self):
+        result = []
+        for one in self:
+            result.append(
+                (one.id, '{} ({:.2f}%)'.format(one.name, one.discount)))
```
Tardo, line 43: I think that here can avoid for loop usage:
```
self.mapped(lambda x: (x.id, '{} ({:.2f}%)'.format(x.name, x.discount)))
```

Tardo (changes requested):

base_global_discount/models/global_discount.py
chienandalu: Doesn't it do it below the surface anyway?

Tardo: Yes, but use mapped it's more "Odoonic"... if tomorrow odoo improve their code you can get the benefits without modify your code. And it's more readable no?

What do you think?

chienandalu: I can buy you the first argument (anyway this model is meant to keep just a handful of registers), but the same could be speculated with the current form.

In the readable part, I'd say the oneliners tend to be more obscure.

Tardo: I see it as syntactic sugar, but of course it's an opinion... totally agree if you don't want use it here.

Tardo (approved):
LGTM :+1:, code review

rafaelbn (approved):
:+1:

PhatDang (commented):
LGTM, code review.

pedrobaeza (approved):

PR #51, opened 2019-11-13
[FIX] base_user_role: Improve tests resiliency

yajo:
These 2 tests were checking the exact set of groups a user should have.

If these tests are ran in a database where a module is previously installed which adds more groups to the base role, these exact group sets would be inexact, although the behavior that is being tested was actually properly working.

With this patch, basically I'm testing if the user contains the groups from the roles, not the exact role set expected. It should work in integration scenarios.

@Tecnativa TT20468

### DIFF
--- a/base_user_role/tests/test_user_role.py
+++ b/base_user_role/tests/test_user_role.py
@@ -98,52 +98,54 @@ class TestUserRole(TransactionCase):
         self.assertEqual(user_group_ids, role_group_ids)
 
     def test_role_unlink(self):
-        # Get role1 groups
-        role1_group_ids = self.role1_id.implied_ids.ids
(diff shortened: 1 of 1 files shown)


pedrobaeza (commented):
I think you should use `assertIn` or `assertNotIn` that are more readable than `<=` kind operations.

yajo:
They are, but they behave differently. `A <= B` checks that _len(A)==0 or A is a subset of B_, while `A in B` checks that _len(A)==1 and is a subset of B_. So, I need `<=`.

pedrobaeza:
OK

sebalix (approved):

pedrobaeza (approved):

PR #52, opened 2019-12-09
[IMP] base_user_role: Adding a new tab for user comments in roles form.

HaraldPanten:
With this new tab, it's possible to describe each role in order to let other users know all details.

### DIFF
--- a/base_user_role/models/role.py
+++ b/base_user_role/models/role.py
@@ -29,6 +29,9 @@ class ResUsersRole(models.Model):
             'base_user_role.ir_module_category_role').id,
         string="Associated category",
         help="Associated group's category")
+    comment = fields.Html(
+        string="Internal Notes",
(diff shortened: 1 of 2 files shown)


sebalix:
Hi @HaraldPanten 
It's not the same commit than https://github.com/OCA/server-backend/pull/39/commits, but I suppose it doesn't matter... ? @pedrobaeza what do you think? Should it be a real cherry-pick from `11.0` branch?

pedrobaeza:
In this case, it doesn't matter, as the author is the same.

sebalix (approved):
Then :+1:

ValentinVinagre (approved):
👍 LGTM

HaraldPanten:
@pedrobaeza Do you that this PR can be merged?

THX!

PR #55, opened 2020-01-06
12.0 fix search with suspend security

mdpena:
Avoid Access Error when search with suspend security.

The Odoo `_search` function performs a  `sudo` on the first line of its
implementation, causing that the uid's wrapper (class BaseSuspendSecurityUid) is lost.

This is evidenced when a function with suspend security is called and in
its implementation needs the values ​​of a 'one2many' field (it does a
`search` in another model without reading access) an Access Error is raised.

### DIFF
--- a/base_suspend_security/models/base.py
+++ b/base_suspend_security/models/base.py
@@ -1,7 +1,7 @@
 # Copyright 2016 ACSONE SA/NV
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from odoo import api, models
+from odoo import api, models, SUPERUSER_ID
 
 from ..base_suspend_security import BaseSuspendSecurityUid
 
@@ -17,3 +17,12 @@ class Base(models.AbstractModel):
                 self.env.cr,
                 BaseSuspendSecurityUid(self.env.uid),
                 self.env.context))
+
+    def sudo(self, user=SUPERUSER_ID):
+        if isinstance(self.env.uid, BaseSuspendSecurityUid):
+            return self.with_env(
+                api.Environment(
+                    self.env.cr, BaseSuspendSecurityUid(user), self.env.context
+                )
+            )
+        return super().sudo(user)
--- a/base_suspend_security/models/base.py
+++ b/base_suspend_security/models/base.py
@@ -19,10 +19,10 @@ class Base(models.AbstractModel):
                 self.env.context))
 
     def sudo(self, user=SUPERUSER_ID):
-        if isinstance(self.env.uid, BaseSuspendSecurityUid):
+        if isinstance(user, BaseSuspendSecurityUid):
             return self.with_env(
                 api.Environment(
-                    self.env.cr, BaseSuspendSecurityUid(user), self.env.context
+                    self.env.cr, user, self.env.context
                 )
             )
         return super().sudo(user)
--- a/base_suspend_security/tests/test_base_suspend_security.py
+++ b/base_suspend_security/tests/test_base_suspend_security.py
@@ -30,3 +30,18 @@ class TestBaseSuspendSecurity(TransactionCase):
         # this tests if _normalize_args conversion works
         self.env['res.users'].browse(
             self.env['res.users'].suspend_security().env.uid)
+
+    def test_suspend_security_on_search(self):
+        user_without_access = self.env["res.users"].create(
+            dict(
+                name="Testing Suspend Security",
+                login="nogroups",
+                email="nogroups@suspendsecurity.com",
+                groups_id=[(5,)],
+            )
+        )
+        # the search is forbidden
+        with self.assertRaises(exceptions.AccessError):
+            self.env["ir.config_parameter"].sudo(user_without_access).search([])
+        # this tests the search
+        self.env["ir.config_parameter"].sudo(user_without_access).suspend_security().search([])
--- a/base_suspend_security/tests/test_base_suspend_security.py
+++ b/base_suspend_security/tests/test_base_suspend_security.py
@@ -40,8 +40,9 @@ class TestBaseSuspendSecurity(TransactionCase):
                 groups_id=[(5,)],
             )
         )
+        model = self.env["ir.config_parameter"]
         # the search is forbidden
         with self.assertRaises(exceptions.AccessError):
-            self.env["ir.config_parameter"].sudo(user_without_access).search([])
+            model.sudo(user_without_access).search([])
         # this tests the search
-        self.env["ir.config_parameter"].sudo(user_without_access).suspend_security().search([])
+            model.sudo(user_without_access).suspend_security().search([])
--- a/base_suspend_security/base_suspend_security.py
+++ b/base_suspend_security/base_suspend_security.py
@@ -4,9 +4,6 @@ from odoo.tools import pycompat
 
 
 class BaseSuspendSecurityUid(int):
-    def __int__(self):
-        return self
-
     def __eq__(self, other):
         if isinstance(other, pycompat.integer_types):
             return False


base_suspend_security/models/base.py
```diff
@@ -17,3 +17,12 @@ def suspend_security(self):
                 self.env.cr,
                 BaseSuspendSecurityUid(self.env.uid),
                 self.env.context))
+
+    def sudo(self, user=SUPERUSER_ID):
+        if isinstance(self.env.uid, BaseSuspendSecurityUid):
```
hbrunn: please add a comment here that this is necessary because of https://github.com/OCA/OCB/blob/12.0/odoo/api.py#L858 which is used in https://github.com/OCA/OCB/blob/12.0/odoo/models.py#L4772

hbrunn: actually, this shouldn't be a problem, as we override __int__: https://github.com/OCA/server-backend/blob/8687422d0eed95a8e0cd2de1b39c621ab1e68c8b/base_suspend_security/base_suspend_security.py#L8 - can you debug where exactly the wrapper is lost?

mdpena: who has the wrapper is the self.uid, when doing sudo the `user` parameter not necessarily has the wrapper but still is important to maintain it and avoid not to break the thread.

mdpena: who has the wrapper is the self.uid, when doing sudo the `user` parameter not necessarily has the wrapper but still is important to maintain it and avoid to break the thread.
Also: 
```
>>> isinstance(int(BaseSuspendSecurityUid(1)), BaseSuspendSecurityUid)
False
```

In python 3.6:  DeprecationWarning: __int__ returned non-int (type BaseSuspendSecurityUid).  The ability to return an instance of a strict subclass of int is deprecated, and may be removed in a future version of Python.

hbrunn: ah, now I see the problem. We will get this issue in other places too, as some of the trickery of this module relies on being able to survive int(). But those we can handle as they come.

Now looking closer at this, I don't believe this is the correct patch. Something like

```python
env['some.model'].suspend_security().sudo(42)
```

must run as user 42 without suspended security. With your code, we'll never be able to unsuspend security.

Is there some other way to achieve what you want?

hbrunn: shouldn't we see the marker in `user` if this comes from a `suspend_security`'d search?

mvaled: > Now looking closer at this, I don't believe this is the correct patch. Something like
> 
> ```python
> env['some.model'].suspend_security().sudo(42)
> ```
> 
> must run as user 42 without suspended security. With your code, we'll never be able to unsuspend security.
> 
> Is there some other way to achieve what you want?

`BaseModel._search` does the following:

```python
        self.sudo(access_rights_uid or self._uid).check_access_rights('read')
```
If the sequence `suspend_security().sudo()` must reactivate the security, maybe the change is to test the argument to `sudo`:

```python
if isinstance(user, BaseSuspendSecurityUid):
   ...
```

mdpena: > ```python
> if isinstance(user, BaseSuspendSecurityUid):
> ```

I agree, that should work.

base_suspend_security/models/base.py
```diff
@@ -17,3 +17,12 @@ def suspend_security(self):
                 self.env.cr,
                 BaseSuspendSecurityUid(self.env.uid),
                 self.env.context))
+
+    def sudo(self, user=SUPERUSER_ID):
+        if isinstance(user, BaseSuspendSecurityUid):
+            return self.with_env(
+                api.Environment(
+                    self.env.cr, BaseSuspendSecurityUid(user), self.env.context
```
hbrunn: we don't need to wrap `user` again

base_suspend_security/tests/test_base_suspend_security.py
```diff
@@ -30,3 +30,25 @@ def test_base_suspend_security(self):
...
+                email="nogroups@suspendsecurity.com",
+                groups_id=[(5,)],
+            )
+        )
+        model_ids = (
+            self.env["ir.model.access"]
+            .search([("group_id", "=", False)])
+            .mapped("model_id.id")
+        )
+        model = self.env["ir.model"].search([("id", "not in", model_ids)], limit=1)
```
hbrunn: I find this a little convoluted. Why not pick some specific model from the base module of which you know it's not accessible?

base_suspend_security/tests/test_base_suspend_security.py
```diff
@@ -30,3 +30,25 @@ def test_base_suspend_security(self):
...
+        # the search is forbidden
+        with self.assertRaises(exceptions.AccessError):
+            self.env[model._name].sudo(user_without_access).search([])
+        # this tests the search
+        self.env[model._name].sudo(user_without_access).suspend_security().search([])
+
```
hbrunn: this line can go away

hbrunn (changes requested):
I added a few nitpicks.

Do you agree we should remove the `__int__` implementation, as it doesn't do what we want anyways?

mdpena:
Is done.

hbrunn (approved):
thanks!

mvaled:
Hi @hbrunn 

Do we require another review?

hbrunn:
yes

mvaled:
> yes

Maybe @pedrobaeza could review this? ;)

pedrobaeza:
Sorry, I'm out of the OCA board this year and stepped out from my community manager role, so you should look for another wildcard.

mvaled:
> Sorry, I'm out of the OCA board this year and stepped out from my community manager role, so you should look for another wildcard.

Sorry to hear about that.  I hope everything goes well for you and your new endeavors.

pedrobaeza:
> Sorry to hear about that. I hope everything goes well for you and your new endeavors.

Not any bad thing. Simply the phase has passed and now others have taken over.

hbrunn:
@mvaled if you approve this, we can merge

mvaled (approved):

mvaled:
I didn't know I could approve it.  I've just learnt how to do it.

hbrunn:
@mvaled do you have a suggestion where to write about that so that more people are ware of this? It would be a pity to lose possible reviewers just because of this kind of misunderstandings.

/ocabot merge patch

hbrunn:
please check https://travis-ci.org/OCA/server-backend/jobs/641058399#L331

mvaled:
> @mvaled do you have a suggestion where to write about that so that more people are ware of this? It would be a pity to lose possible reviewers just because of this kind of misunderstandings.

Since I'm not a regular contributor to OCA projects I simply assumed I wasn't allowed.  I guess the README could point to general OCA guidelines about contributions and reviews.  Getting a formal statement of how do people get promoted from 'regular guy who reports and fixes some issues once in a while' to reviewer would help.  Specially for those seeking to get really involved the OCA, if they happen to land in any repository, a link to OCA's Get Involved page could help.

(I just noticed this https://odoo-community.org/page/review, so it seems anyone can submit a review. Cool!)

hbrunn:
ah, then the omittance is mainly github stuff. you're allowed to review the hell out of all public repos here. rules here are we need to two approvals, but need someone with push permissions approve, and that would be me. and days in between, that's also agreed on for others to intervene if necessary.

but now the one thing that makes the merge bot fail can be fixed easily, and then we're done

mvaled:
> but now the one thing that makes the merge bot fail can be fixed easily, and then we're done

I talked to @mdpena so that she works this out.

mdpena:
@hbrunn @mvaled 

Is done. Thanks, both.

PR #61, opened 2020-01-12
[11.0] Add base_dav

fkantelberg:
Hello,

this PR continues the work of @hbrunn of #26 . 

The following configuration should provide a testing collection using the odoo calendar allowing to test the calendar using an email client:
grafik


robyf70:
@fkantelberg I'm looking at this module to see if I can get working an WebCal for calendar sharing support under Odoo v12. I would like to share Odoo calendar among smartphone, thunderbird etc etc in order to get automatic contact's notification. Do you have an example that works with Odoo calendar?

fkantelberg:
@robyf70 Sadly this version isn't ported to v12 and you should use odoo 11 for testing.

See the image in the initial post here for the odoo side configuration. Additional it's recommended to create a UUID field for calender.event in odoo and link it otherwise it'll cause trouble if you create events in your app and sync it to odoo. Until now it seems to work with thunderbird and basic calendar apps but it has some trouble with more extended calendar apps which want to manage multiple calendars under one URL.

robyf70:
@fkantelberg Ok! I've ported it to v12 with some changes. I will do some tests as you suggested

OSevangelist:
@robyf70 would you mind to contribute your port to v12 back ? Would be appreciated

gurneyalex:
Hello I just triggered a rebuild on runbot and I get a red build with this stack trace:

```
File "/home/odoo/build/OCA/server-backend/base_dav/models/__init__.py", line 3, in <module>
from . import dav_collection
File "/home/odoo/build/OCA/server-backend/base_dav/models/dav_collection.py", line 15, in <module>
from ..radicale.collection import Collection, FileItem, Item
File "/home/odoo/build/OCA/server-backend/base_dav/radicale/__init__.py", line 4, in <module>
from . import collection
File "/home/odoo/build/OCA/server-backend/base_dav/radicale/collection.py", line 26, in <module>
class FileItem(Item):
TypeError: NoneType takes no arguments
```

gurneyalex:
PR https://github.com/initOS/server-backend/pull/3 fixes the build by preventing the installation of radicale >= 3.0.0

fkantelberg:
@gurneyalex Thank you for the fix. I merged it and the runbot is up.

gurneyalex:
a colleague (@yterrettaz)  tested the PR on runbot and gets a crash:

bug_webdav

fkantelberg:
@gurneyalex The last patch fixed that problem. For a calendar I highly recommend to set the field UUID.

base_dav/views/dav_collection.xml
```diff
@@ -0,0 +1,77 @@
...
+            </tree>
+        </field>
+    </record>
+
+    <record id="view_dav_collection_mapping_form" model="ir.ui.view">
+        <field name="model">dav.collection.field_mapping</field>
+        <field name="arch" type="xml">
+            <form>
+                <group>
+                    <field name="collection_id" invisible="1"/>
```
yvaucher, line 53: When trying to create a field mapping from the view, we got

```
The following fields are invalid:

Collection
```

Default value must be set on parent view.

Or if that view is only used embedded in the collection view you can try to remove this field completely. Not sure it would work with a Form view but Odoo is sets automatically the field like on `account.invoice.line` without the default but it's an inline edition with a tree view. There are some magic behind that and I don't know on which conditions it might not work with Form view, thus probably safer to set the default in context as in `res.partner`.

base_dav/views/dav_collection.xml
```diff
@@ -0,0 +1,77 @@
...
+                </group>
+                <group string="Access">
+                    <field name="model_id"/>
+                    <!-- TODO: use widget="domain" when we can set the model dynamically /-->
+                    <field name="domain"/>
+                    <field name="field_uuid" domain="[('model_id', '=', model_id)]"/>
+                    <field name="rights"/>
+                </group>
+                <group string="Additional field mapping">
+                    <field name="field_mapping_ids" nolabel="1"/>
```
yvaucher: To set the default value to link correctly a field mapping from the UI:

```suggestion
                    <field name="field_mapping_ids" nolabel="1" context="{'default_collection_id': active_id}/>
```

yvaucher (changes requested):
It's impossible to create a field mapping from the UI an error. (tested on run bot)

error_collection

(an other screen capture by @yterrettaz )



It needs a default value on `collection_id`

base_dav/views/dav_collection.xml
yvaucher: (BTW I did this change by hand on http://3432983-61-2e5e17.runbot2.odoo-community.org/web?debug=1#id=2&view_type=form&model=dav.collection&action=116 )

It fixes it.

hbrunn:
FYI I abandoned the *dav route completely and am looking into creating `base_activesync` with does more or less the same as this, but then using [ActiveSync](https://docs.microsoft.com/en-us/openspecs/exchange_server_protocols/MS-OXPROTLP/30c90a39-9adf-472b-8b5b-03c282304a83) as you nearly don't find a mobile device not supporting it.

This will take *long* as there's no ready made server library for this in python, which I'll have to create first. Funding can speed this up of course.

gurneyalex:
> @gurneyalex The last patch fixed that problem. For a calendar I highly recommend to set the field UUID.

@fkantelberg  any suggestion about what to put in there, what this is used for? Maybe a note in the README  / a help on the field would be useful...

fkantelberg:
Thanks @yvaucher

@gurneyalex 
Sadly there is no good field at the calender.event. Normally I'm defining one in a module which builds on top of this.

```
import uuid

from odoo import fields, models

class CalenderEvent(models.Model):
    _inherit = "calendar.event"

    uuid = fields.Char(default=lambda self: str(uuid.uuid4().hex))
```

I agree that this is really annoying to test now but I don't think that this is part of the base_dav module. I would suggest to make a simple module (calendar_dav?) which automatically installs when calendar and base_dav are installed but I'm not sure in which repository should it be placed.

yvaucher (approved):

Rad0van:
Could this be used to provide WebDav backend to attachments and maybe allow (with cooperation with new [DMS module](https://github.com/OCA/dms) ) allow viewing/editing attachments (i.e. Word / Excel documents) without the need to download them locally and the reupload them?

What is the status of this? Has anyone tackled this for 13.0?

fkantelberg:
I haven't really tested the file management. The upload and deletion isn't implemented for attachments because my focus was to make the CalDav part work in the first iteration but the module should be able to handle it after the functions are properly implemented.

I might have a working v12 version but have to check if the latest changes are also covered.

s1risaac:
> I haven't really tested the file management. The upload and deletion isn't implemented for attachments because my focus was to make the CalDav part work in the first iteration but the module should be able to handle it after the functions are properly implemented.
> 
> I might have a working v12 version but have to check if the latest changes are also covered.

Is there a progress to insert CalDav to Version 12?

houssine78 (approved):
Code LGTM, no test

fkantelberg:
@gurneyalex would you merge it?

jpoa:
Greetings, any chance of this getting through?

bosd:
> Is there a progress to insert CalDav to Version 12?

@fkantelberg Is there a port available to a recent odoo version?

PR #64, opened 2020-02-25
[FIX] base_user_role: default user multi-company

oerp-odoo:
When creating user, default roles are using `base.default_user`, which
in multi-company case, can be from different company than user is being
created for. If thats the case, user creating another user, will get
access error, when trying to read default template user.

To work around that, we use sudo, to make sure, template user data can
be read regardless of a company.

Also improved the way base_user_role tests are run. Changed base test
class from TransactionCase to SavepointCase.

From how tests are ran perspective, both classes behave the same: each
test uses set up which is roll backed after.

But the difference is that SavepointCase case is much faster, because
it calls setUpClass once and saves this case for all tests (reusing
saved state).

NOTE. SavepointCase for these particular tests, increases tests speed (at least in my case) more than 2 times. I highly recommend using `SavepointCase` for almost all tests classes, instead of `TransactionCase`, especially when set up is big (like need to create/update lots of records etc).

### DIFF
--- a/base_user_role/README.rst
+++ b/base_user_role/README.rst
@@ -23,7 +23,7 @@ User roles
     :target: https://runbot.odoo-community.org/runbot/253/12.0
     :alt: Try me on Runbot
 
-|badge1| |badge2| |badge3| |badge4| |badge5| 
+|badge1| |badge2| |badge3| |badge4| |badge5|
 
 This module was written to extend the standard functionality regarding users
 and groups management.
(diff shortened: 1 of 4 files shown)


oerp-odoo:
Also note, there is another issue with multi-company and user access.
If you create Role and assign some users to that role. Then create another company, switch to that company and try to create/assign another user for that role, you will get access denied error.

It is because `_compute_user_ids` `update_users` methods, will try to scan all users using current user access rights - that is trying to update/change role. And if user is not superuser, he won't be able to save changes (normal user does not have access to mulitple companies at once. Unless those companies are childs to his company, he currently represents).

But I dont know if this module is supposed to handle multi-company roles. I mean if one role is intended to be used for users from multiple companies? Or there should be different roles for that?

Though, then there might be issue when users switch companies and different roles become incorrect..

StefanRijnhart (approved):
Makes sense! And +1 for improving the performance of the test. Can you rebase?

oerp-odoo:
@StefanRijnhart rebased

legalsylvain (approved):

PR #65, opened 2020-02-25
[ADD][12.0] server_action_navigate

legalsylvain:
Supersed : https://github.com/OCA/server-ux/pull/129

This module allow to configure new server Actions to navigate between models.

for exemple, if you want to have the list of the products, that has been sold in some sale orders, you can configure a new Server Actions like this : 

ir_actions_server_form

then you'll have a new option in the "Action" list : 

sale_order_tree

Once the button clicked, the result will be a list of products : 

product_product_tree

Technically, it adds a new selection key ``navigate`` in the field ``state`` of the model ``ir.actions.server``.

CC : @hbrunn, @pedrobaeza, @elvise

#GRAPOCA

PR #70, opened 2020-03-16
[12.0][ADD] new module server_action_sort to sort lines of any models by any criterias

legalsylvain:
Add new module ``server_action_sort``. (similar to ``server_action_navigate`` available [here ](https://github.com/OCA/server-backend/pull/65))
Replace previous module ``mass_sorting`` in server-tools repository for V7 / V8 and [V10](https://github.com/OCA/server-tools/tree/10.0/mass_sorting)

This module allow to configure new server Actions to sort items lines.

For exemple, if you want to sort sale order lines, by any criterias, you can define a new server action like this : 

ir_actions_server_form

A new option, is then available in the action menu :

sale_order_tree

This module provides a generic way to do the same things than other modules like : 

- ``purchase_order_line_deep_sort`` (purchase-workflow [V12](https://github.com/OCA/purchase-workflow/tree/12.0/purchase_order_line_deep_sort)) CC : @cubells, @sergio-teruel 
- ``account_invoice_line_sort`` (account-invoicing) CC : @Cedric-Pigeon


pedrobaeza:
There are some unrelated commits here. I prefer the name `mass_sorting` :stuck_out_tongue_closed_eyes:

legalsylvain:
> There are some unrelated commits here.

Fixed.

> I prefer the name mass_sorting

Well, the features is the same, but the design has been totally reworked. It is now based on server_action exactly like ``server_action_navigate``. [Ref](https://github.com/OCA/server-ux/pull/129#issuecomment-590908689). I think it's great if all the modules based on the same concept have the same syntax for the technical name. But for the "friendly name", I set "Server Actions - Mass Sort Lines". 
Is it OK for you ?

pedrobaeza:
Yeah, good, I only express my feeling, but the author decides. You can also call it the "sort-enator", hahaha

legalsylvain:
Hi @dreispt, @hbrunn, @flotho could you take a look on that PR (similar to ``server_action_navigate`` that is been merged here https://github.com/OCA/server-backend/pull/65)

It is working on production since 6 monthes without problem.

PR #74, opened 2020-05-18
[13.0][MIG]base_import_match

chusamo:
I just changed the versio number, and checked that works well.


pedrobaeza:
Please preserve commit history following technical method explained in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-13.0

PR #77, opened 2020-06-02
[FIX] base_user_role: ensure Channel autosubscription logic is applied

dreispt:
It could be the case where ``mail`` Channel autosubscription logic would
run before the Role logic added new groups to the User.
In that case, the autosubscrition logic would fail to apply.

Adding the ``mail`` dependency ensures this does not happen.

While the ``mail`` module in not required for the Role features,
it is a reasonable dependency to add, since it will be already
present in the vast majority of Odoo installations.


dreispt:
Ping!

StefanRijnhart (approved):

PR #83, opened 2020-08-04
[12.0][ADD] fetchmail_local_mailbox

hbrunn:
This module allows to retrieve mails from a mailbox on the local file system.

That's useful when your Odoo instance runs on a server with a local mailserver that delivers mails into a mailbox accessible by the user Odoo is running as, because it allows you to fetch mails without any credentials or TCP/connection overhead.


fetchmail_local_mailbox/models/fetchmail_server.py
```diff
@@ -0,0 +1,108 @@
...
+
+_logger = logging.getLogger(__name__)
+
+
+class FetchmailServer(models.Model):
+    _inherit = "fetchmail.server"
+
+    type = fields.Selection(selection_add=[("local_mailbox", "Local mailbox")])
+    mailbox_type = fields.Selection([("maildir", "Maildir")], default="maildir")
+    mailbox_path = fields.Char()
```
NL66278, line 17: Should this not be required?

PR #89, opened 2020-08-27
[WIP][13.0] add base_external_system_ftp DO NOT REVIEW OR MERGE YET!

NL66278:


rvalyi:
Hello @NL66278 I didn't check a lot, but may be there is some overlap with this module we contributed recently https://github.com/OCA/server-tools/tree/12.0/attachment_synchronize Just worth double checking.

NL66278:
@rvalyi While there is some overlap and attachment_synchronize is definitely a nice module, server_ftp is intended to be much more simple to setup on the one hand, and to profit from a centralized definition of external systems on the other hand. That is why we rejected basing our module on storage_backend_sftp as well.

rvalyi:
@NL66278 no problem with that personally, I just wanted to make sure you were aware of it and things don't get duplicated just because of lack of information. Good luck with the PR.

NL66278:
Codecov complains about server interfaces that are not tested, because these are for external systems. However all basecode is fully covered, and the principles behind the remote systems are tested with a mock server...

PR #90, opened 2020-09-01
[13.0][MIG] base_external_dbsource_mysql

dalonsod:
Standard migration


dalonsod:
@pedrobaeza @sergio-teruel I'm not sure how codecov warnings could be attended, could you help me?

Also, I've got two doubts about this migration:
- with external python dependencies properly defined in manifest file (according v13 migration instructions, https://github.com/odoo/odoo/pull/25549), is it still required import validation code developed in `models/base_external_dbsource.py` file?
- is `sqlalchemy` external dependency still necessary if it's already defined in `base_external_dbsource` addon?

Thanks!

base_external_dbsource_mysql/models/base_external_dbsource.py
```diff
@@ -0,0 +1,58 @@
...
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+
+import logging
+
+from odoo import models
+
+_logger = logging.getLogger(__name__)
+
+try:
+    from odoo.addons.base_external_dbsource.models import base_external_dbsource
```
NL66278: Why is this line in the try block?

base_external_dbsource_mysql/models/base_external_dbsource.py
```diff
@@ -0,0 +1,58 @@
...
+        )
+    try:
+        import sqlalchemy
+    except ImportError:
+        _logger.info(
+            "SQLAlchemy library not available. Please "
+            'install "sqlalchemy" python package.'
+        )
+except ImportError:
+    _logger.info("base_external_dbsource Odoo module not found.")
```
NL66278: This can not happen, as you have it in the depends of this module. If module you depend on is not there, your module will not load and this line will never be reached.

base_external_dbsource_mysql/models/base_external_dbsource.py
```diff
@@ -0,0 +1,58 @@
...
+_logger = logging.getLogger(__name__)
+
+try:
+    from odoo.addons.base_external_dbsource.models import base_external_dbsource
+
+    CONNECTORS = base_external_dbsource.BaseExternalDbsource.CONNECTORS
+    try:
+        import MySQLdb
+
+        CONNECTORS.append(("mysql", "MySQL"))
```
NL66278: Why not handle this exactly like the sqlalchemy import, and do away with the assert?

pedrobaeza:
Codecov errors comes because the total test coverage is lower, and that means this module doesn't have enough tests for the code that is adding. It's not mandatory to add such tests, although advisable. This being a "connector" to other databases, it's difficult though to create tests, and doing mocks for simulating inputs is not guarantee for proper working of the module.

NL66278 (changes requested):
See other comments. Also need to increase test coverage.

dalonsod:
> 
> 
> See other comments. Also need to increase test coverage.

Thanks @NL66278 for your review. I agree with your comments, the related code came from v12 implementation, and that's why I've asked before if all import validations are still required. Maybe the code can be simplified and only do the imports and connector append stuff

dalonsod:
> 
> 
> Codecov errors comes because the total test coverage is lower, and that means this module doesn't have enough tests for the code that is adding. It's not mandatory to add such tests, although advisable. This being a "connector" to other databases, it's difficult though to create tests, and doing mocks for simulating inputs is not guarantee for proper working of the module.

Thanks, I'll ignore them.

NL66278:
@dalonsod Yes, please do some simplifications.

dalonsod:
> @dalonsod Yes, please do some simplifications.

Done, even MySQL import seems unnecessary, locally tested

NL66278 (approved):
:+1: Thank you for the migration! LGTM

dalonsod:
@sergio-teruel could you review? Thanks!

dalonsod:
@pedrobaeza @AdriaGForgeFlow could you review this? Thanks in advance

pedrobaeza (approved):

Issue #91, opened 2020-09-28
[13.0] base_user_role : Group default application is not set on role creation.


eferriere:
Impacted versions: 13.0
Module: base_user_role

Steps to reproduce:

* Create a role from the role menu point, filling the name it will be given.

Current behavior:

* When a role is created, no application is set on the user group, which has been created with the role and is linked to it.

Expected behavior:

* On role creation, a user group must be created linked to the role, with the "User roles" category (XML ID: "base_user_role.ir_module_category_role") as application.

eferriere:
This issue is linked to the #92 pull request.

StefanRijnhart:
Closing as the fix is being merged. Thanks for reporting and fixing the bug! If you plan to file a PR, then next time you don't have to open a separate issue.

PR #92, opened 2020-09-28
[13.0] base_user_role: Corrected a bug on the attribution of the 'User roles' default applic…

eferriere:
…ation to a created user group on role creation.

### DIFF
--- a/base_user_role/models/role.py
+++ b/base_user_role/models/role.py
@@ -32,6 +32,7 @@ class ResUsersRole(models.Model):
         default=lambda cls: cls.env.ref("base_user_role.ir_module_category_role").id,
         string="Associated category",
         help="Associated group's category",
+        readonly=False,
     )
 
     @api.depends("line_ids.user_id")


vincent-hatakeyama:
This is a fix for #91

base_user_role/models/role.py
```diff
@@ -27,12 +38,6 @@ class ResUsersRole(models.Model):
     user_ids = fields.One2many(
         comodel_name="res.users", string="Users list", compute="_compute_user_ids"
     )
-    group_category_id = fields.Many2one(
-        related="group_id.category_id",
-        default=lambda cls: cls.env.ref("base_user_role.ir_module_category_role").id,
-        string="Associated category",
-        help="Associated group's category",
-    )
```
sebalix, line 35: This code was working well before, so I suppose the issue is that the default value is not set anymore because starting from Odoo 12.0 the related fields are `readonly` by default. Try with `readonly=False` here and check if it solves the issue?

eferriere: Setting the "readonly" attribute to "False" indeed works.

base_user_role/models/role.py
```diff
@@ -27,11 +27,13 @@ class ResUsersRole(models.Model):
     user_ids = fields.One2many(
         comodel_name="res.users", string="Users list", compute="_compute_user_ids"
     )
+
```
sebalix: Can you remove this empty line to reduce the diff?

sebalix (commented):
Minor comment, thanks for the fix!

It'll need to be backported in 12.0, if you want to do it it'll be very nice, otherwise I'll do it myself later :)

eferriere:
Thank you for the revision.

Unfortunately, I will not be able to port to Odoo 12 these modifications.

sebalix (approved):

alan196 (approved):

sebalix:
Commits needs to be squashed before merging this one.

StefanRijnhart (approved):

PR #94, opened 2020-10-07
[FIX] do not raise an error when default_user is not accessible by ir.rule

legalsylvain:
Trivial Fix. 

when creating a user, we want to access to the role of the default user. if the user is not readable (due to ir.rule), the creation fails. This commit fixes the problem.

### DIFF
--- a/base_user_role/models/user.py
+++ b/base_user_role/models/user.py
@@ -25,7 +25,7 @@ class ResUsers(models.Model):
         )
         default_values = []
         if default_user:
-            for role_line in default_user.role_line_ids:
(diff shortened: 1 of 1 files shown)


StefanRijnhart:
For the record, this is a duplicate of https://github.com/OCA/server-backend/pull/64/files#diff-446dbf60f9fb943a77d84e50d6f2bf695700fe5192986cd65bd23fb10701e061R24

legalsylvain:
Thanks @StefanRijnhart, I didn't saw that PR.
closing as #64 is  being merged

PR #96, opened 2020-11-04
[13.0][MIG] base_global_discount: Migration to 13.0

ernestotejeda:
cc @Tecnativa TT26603

PR #97, opened 2020-11-04
[13.0][FIX] base_user_role: Improve tests resiliency

ernestotejeda:
cc @Tecnativa TT26603
Needed for https://github.com/OCA/server-backend/pull/96

I am applying here a commit that was made for this module in v12.0 but wasn't applied in v13.0. If this is not applied in v13.0, the tests of this module may fail under certain conditions.


pedrobaeza:
As this is a cherry pick, I'm fast tracking it:

/ocabot merge patch

PR #102, opened 2020-12-11
[IMP] base_global_discount: add specific security group

chienandalu:
This way, it will be easier to decide who can set discount or not. The
current group scope is too broad.

cc @Tecnativa TT26740

### DIFF
--- a/base_global_discount/__manifest__.py
+++ b/base_global_discount/__manifest__.py
@@ -13,8 +13,8 @@
         'product',
     ],
     'data': [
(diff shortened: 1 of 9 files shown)


HaraldPanten (approved):

pedrobaeza (approved):

pedrobaeza:
/ocabot merge major

Please forward-port it to v13

StefanRijnhart:
I managed to fix three bugs in this change here: https://github.com/OCA/server-backend/pull/103

PR #106, opened 2020-12-20
[IMP] base_user_role: Adding a new tab for user comments in roles form.

HaraldPanten:
As it is in V12.

It's useful as many times users like to have a summary of the rules, or any other changes.

### DIFF
--- a/base_user_role/views/role.xml
+++ b/base_user_role/views/role.xml
@@ -34,6 +34,9 @@
                                 </tree>
                             </field>
                         </page>
+                        <page name="page_internal_notes" string="Internal Notes">
+                            <field name="comment" />
+                        </page>
                     </notebook>
                 </sheet>
             </form>
--- a/base_user_role/models/role.py
+++ b/base_user_role/models/role.py
@@ -34,6 +34,7 @@ class ResUsersRole(models.Model):
         help="Associated group's category",
         readonly=False,
     )
+    comment = fields.Html(string="Internal Notes",)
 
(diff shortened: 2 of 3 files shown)


HaraldPanten:
@pedrobaeza do you mind merging that easy one, to keep consistancy in newer versions? By the way, should I open another PR for V14?

THX

base_user_role/views/role.xml
```diff
@@ -34,6 +34,9 @@
                                 </tree>
                             </field>
                         </page>
+                        <page name="role_inernal_notes" string="Internal Notes">
```
pedrobaeza: Typo and `role` is redundant. Better to prefix it with the type of element:

```suggestion
                        <page name="page_internal_notes" string="Internal Notes">
```

HaraldPanten:
@pedrobaeza done

base_user_role/views/role.xml
```diff
@@ -34,6 +34,9 @@
                                 </tree>
                             </field>
                         </page>
+                        <page name="page_inernal_notes" string="Internal Notes">
```
pedrobaeza:
```suggestion
                        <page name="page_internal_notes" string="Internal Notes">
```

HaraldPanten:
OMG... sorry Pedro.

You know that "haste makes waste"!

pedrobaeza (approved):

HaraldPanten:
Should I do that for V14 as well?

PR #108, opened 2021-01-25
[FIX] base_user_role: remove multicompany functionality

kevinkhao:
Multicompany functionality is not the same in 14.0.

To illustrate:

Screenshot from 2021-01-25 22-53-46

Before, selecting one company or the other would change the user's company_id (in DB). Now, it's only a key in context.

So current multicompany functionality is broken and to be re-imagined later. This PR removes multicompany functionality.

### DIFF
--- a/base_user_role/views/user.xml
+++ b/base_user_role/views/user.xml
@@ -16,10 +16,6 @@
                             <field name="date_from" />
                             <field name="date_to" />
                             <field name="is_enabled" />
-                            <field
-                                name="company_id"
-                                groups="base.group_multi_company"
-                            />
                         </tree>
                     </field>
                 </page>
--- a/base_user_role/models/role.py
+++ b/base_user_role/models/role.py
@@ -3,8 +3,7 @@
 import datetime
 import logging
 
-from odoo import SUPERUSER_ID, _, api, fields, models
-from odoo.exceptions import ValidationError
+from odoo import SUPERUSER_ID, api, fields, models
 
 _logger = logging.getLogger(__name__)
 
@@ -89,23 +88,6 @@ class ResUsersRoleLine(models.Model):
     date_from = fields.Date("From")
     date_to = fields.Date("To")
     is_enabled = fields.Boolean("Enabled", compute="_compute_is_enabled")
-    company_id = fields.Many2one(
-        "res.company", "Company", default=lambda self: self.env.user.company_id
-    )
-
-    @api.constrains("user_id", "company_id")
-    def _check_company(self):
-        for record in self:
-            if (
-                record.company_id
-                and record.company_id != record.user_id.company_id
-                and record.company_id not in record.user_id.company_ids
-            ):
-                raise ValidationError(
-                    _('User "{}" does not have access to the company "{}"').format(
-                        record.user_id.name, record.company_id.name
-                    )
-                )
 
     @api.depends("date_from", "date_to")
     def _compute_is_enabled(self):
--- a/base_user_role/models/user.py
+++ b/base_user_role/models/user.py
@@ -49,10 +49,7 @@ class ResUsers(models.Model):
         return res
 
     def _get_enabled_roles(self):
-        return self.role_line_ids.filtered(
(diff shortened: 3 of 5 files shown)


base_user_role/views/user.xml
```diff
@@ -16,24 +16,10 @@
...
-                                name="company_id"
-                                groups="base.group_multi_company"
-                            />
                         </tree>
                     </field>
                 </page>
             </xpath>
         </field>
     </record>
-    <record id="view_res_users_search_inherit" model="ir.ui.view">
```
sebastienbeau: Why do you remove this search?

kevinkhao: mistake. fixed

sebastienbeau (approved):
LGTM

sebastienbeau:
@PierrickBrun

PierrickBrun (approved):

CasVissers-360ERP:
@sebastienbeau Isn't this a bit bluntly? Just removing something? Customer are depending on this module with multi-company features. A migration should be for the module in totality? 

cc @dreispt

CasVissers-360ERP:
In addition on my installation it blocks me from updating the module:
"views/role.xml", should be loaded before "data/ir_module_category.xml",

sebastienbeau:
> @sebastienbeau Isn't this a bit bluntly? Just removing something? Customer are depending on this module with multi-company features. A migration should be for the module in totality? 
> 
> cc @dreispt 


Sadly due to odoo multicompany change of implementation since v13. It impossible to implement the same behaviour (user can be in several companies at the same time, so having different right/menu is a nightmare), this is why we drop it. We are the initial contributor of this feature so we didn't remove it by pleasure.

Now our customer just create different user when he need different right per company.

In any case if you found a way to reimplement this feature we are interested but this should be done in a separated module (as we need to change odoo behaviour on multicompany).

sebastienbeau:
> In addition on my installation it blocks me from updating the module:
> "views/role.xml", should be loaded before "data/ir_module_category.xml",

@kevinkhao can you check this point

jcdrubay:
I have been using base_user_role on several projects since version 10, but over the time, I found that for most of projects, `base_user_role` was over-engineered, and then it just became more and more complex. Then, due to this complexity, it seems more and more difficult to port to latest version of Odoo.

Therefore, I see this removal as a good change and I hope we could simplify even more the module base_user_role and split non-core features to extra modules.

For the time being, we (at Komit) decided to create and use a different module `user_profile` which adds a m2m field profile_ids. Basically, it behaves the same as base_user_role when only setting roles without dates, enable/disable, company on the roles of a user in a user form.

@sebastienbeau I would be happy to contribute it back to OCA (the code was inspired from base_user_role). Or do you think we should avoid modules competing each other on similar features within OCA? Or do you think it could be part of the Roadmap to split `base_user_role` into smaller modules?

sebastienbeau:
@jcdrubay interesting feedback. On our side we do not use the enable/date feature. And a simple m2m for most of the case will improve the UI.
It will be interesting to have more feedback. Can you open an issue to propose to split this module in two? base_user_role and base_user_role_date?

kevinkhao:
Couldn't reproduce the issue by installing + testing on a fresh DB. Maybe a migration issue ?

CasVissers-360ERP:
> > @sebastienbeau Isn't this a bit bluntly? Just removing something? Customer are depending on this module with multi-company features. A migration should be for the module in totality?
> > cc @dreispt
> 
> Sadly due to odoo multicompany change of implementation since v13. It impossible to implement the same behaviour (user can be in several companies at the same time, so having different right/menu is a nightmare), this is why we drop it. We are the initial contributor of this feature so we didn't remove it by pleasure.
> 
> Now our customer just create different user when he need different right per company.
> 
> In any case if you found a way to reimplement this feature we are interested but this should be done in a separated module (as we need to change odoo behaviour on multicompany).

Ok, thank you for the clear explanation. It might be an option to see if a user is logged into one or multiple companies from the context, if just one, load the related role lines, else load role lines where company is False?

PR #109, opened 2021-01-25
[IMP] base_user_role: UX improvements

kevinkhao:
Roles should not be able to be picked several times for a user. Show only relevant roles on user role lines.

Screenshot from 2021-01-26 00-34-57

Screenshot from 2021-01-26 00-35-18

### DIFF
--- a/base_user_role/views/user.xml
+++ b/base_user_role/views/user.xml
@@ -10,9 +10,13 @@
         <field name="arch" type="xml">
             <xpath expr="//notebook/page[1]" position="before">
                 <page string="Roles">
+                    <field name="role_ids" invisible="1" />
                     <field name="role_line_ids" nolabel="1">
                         <tree editable="bottom" decoration-muted="not is_enabled">
-                            <field name="role_id" />
+                            <field
+                                name="role_id"
+                                domain="[('id', 'not in', parent.role_ids)]"
+                            />
                             <field name="date_from" />
                             <field name="date_to" />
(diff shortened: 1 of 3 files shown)


yostashiro:
Functional test. LGTM 👍

kevinkhao:
ping @sebastienbeau

base_user_role/views/user.xml
```diff
@@ -12,7 +12,11 @@
                 <page string="Roles">
                     <field name="role_line_ids" nolabel="1">
                         <tree editable="bottom" decoration-muted="not is_enabled">
-                            <field name="role_id" />
+                            <field name="user_roles_already_picked" invisible="1" />
+                            <field
+                                name="role_id"
+                                domain="[('id', 'not in', user_roles_already_picked)]"
```
sebastienbeau: you can make it simplier.
There is already a computed field on the res.user see here https://github.com/OCA/server-backend/blob/14.0/base_user_role/models/user.py#L15

You can simply add this field in the view (in invisible) and then use the domain 
 domain="[('id', 'not in', parent.role_ids)]"

kevinkhao: changed + functional tested

sebastienbeau:
@kevinkhao can you fix the lint?

kevinkhao:
@sebastienbeau done

sebastienbeau (approved):
LGTM code review

PR #111, opened 2021-02-12
[12.0][FIX] base_suspend_security: Pretend to be a list of ints if neccessary

hbrunn:
when searching one2many fields, our uid ends up in the code branch for [iterables](https://github.com/OCA/OCB/blob/12.0/odoo/osv/expression.py#L973), so the uid needs to support [len and slicing](https://github.com/OCA/OCB/blob/12.0/odoo/osv/expression.py#L448)

The second commit is a forward port of a quite crucial [v8 feature](https://github.com/OCA/server-tools/pull/969), described in depth [later in the thread](https://github.com/OCA/server-tools/pull/969#issuecomment-330002605)

### DIFF
--- a/base_suspend_security/tests/test_base_suspend_security.py
+++ b/base_suspend_security/tests/test_base_suspend_security.py
@@ -55,7 +55,7 @@ class TestBaseSuspendSecurity(TransactionCase):
         with self.assertRaises(exceptions.AccessError):
             model.sudo(user_without_access).search([])
         # this tests the search
-            model.sudo(user_without_access).suspend_security().search([])
+        model.sudo(user_without_access).suspend_security().search([])
         # be sure we can search suspended uids like ints
         partners = self.env['res.partner'].with_context(
             active_test=False,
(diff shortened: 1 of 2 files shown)


base_suspend_security/tests/test_base_suspend_security.py
```diff
@@ -46,3 +56,20 @@ def test_suspend_security_on_search(self):
             model.sudo(user_without_access).search([])
         # this tests the search
             model.sudo(user_without_access).suspend_security().search([])
```
astirpe: Isn't this line indented "too much"?

```suggestion
        model.sudo(user_without_access).suspend_security().search([])
```

hbrunn: thanks, I always strife for minimal diffs, but fixing this on the way doesn't hurt

astirpe (approved):

StefanRijnhart (approved):

PR #113, opened 2021-03-15
 [FIX] base_user_role: Do not drop groups, if user has no role_line_ids

legalsylvain:
Proposing a patch that makes the module ``base_user_role`` safer IMO.

Step to reproduce : 
- Create role with groups
- affect a role to a user : the groups will be added / removed, to fit to the role.

before the PR : 
- remove the role : all the groups will be dropped.
it is so not possible to remove roles to "handle manually groups".

after the PR : 
- remove the role : the groups will be conserved.

Note : the PR changes the behaviour only if users doesn't have any roles, making the code consistent with that line :  https://github.com/OCA/server-backend/blob/12.0/base_user_role/models/user.py#L83

CC : @kevinkhao, @sebalix

### DIFF
--- a/base_user_role/models/role.py
+++ b/base_user_role/models/role.py
@@ -112,6 +112,6 @@ class ResUsersRoleLine(models.Model):
     @api.multi
     def unlink(self):
         users = self.mapped("user_id")
-        res = super(ResUsersRoleLine, self).unlink()
-        users.set_groups_from_roles(force=True)
+        res = super().unlink()
+        users.filtered(lambda x: x.role_line_ids).set_groups_from_roles(force=True)
         return res
--- a/base_user_role/tests/test_user_role.py
+++ b/base_user_role/tests/test_user_role.py
@@ -192,8 +192,8 @@ class TestUserRole(TransactionCase):
         self.user_id.role_line_ids.filtered(
             lambda l: l.role_id.id == self.role1_id.id
         ).unlink()
-        # Check user has no groups from role1 and role2
-        self.assertFalse(role1_groups <= self.user_id.groups_id)
+        # Keep user with the groups of role1
+        self.assertLessEqual(role1_groups, self.user_id.groups_id)
         self.assertFalse(role2_groups <= self.user_id.groups_id)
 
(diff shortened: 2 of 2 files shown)


legalsylvain:
I don't think such change is acceptable (that it is the right design)

However, droping all roles should not drop all groups in my opinion.
Specially the internal type of the user, that let the user in a instable state.

What are your opinion ?

kevinkhao:
If we do such an exception (no roles), then I think inevitably we need to manage both user roles and traditional group-based rights at the same time.

Maybe this could be solved with the possibility of having both user roles and traditional group-based rights with a flag ?
Or, we put a big flashing red warning sign saying that this module will turn all permissions upside down and it is the integrator's responsibility to ensure the module is fully configured.

I don't see an easy fix for this issue.

chusamo (approved):
https://www.loom.com/share/3489f9a06ba54af3998e3940587601e6

What I'm seeing this PR will conserve the groups of the last role unlinked. Could be useful for saving time, however as @kevinkhao we have to notify the users because could be a non expected behaivor with wrong configuration for handling manually the groups

Functional Review LGtM

base_user_role/models/role.py
```diff
@@ -112,6 +112,6 @@ def _compute_is_enabled(self):
     @api.multi
     def unlink(self):
         users = self.mapped("user_id")
-        res = super(ResUsersRoleLine, self).unlink()
-        users.set_groups_from_roles(force=True)
+        res = super().unlink()
+        users.filtered(lambda x: x.role_line_ids).set_groups_from_roles(force=True)
```
sebalix, line 116: Sorry for this late review...

The `force=True` parameter has been added on purpose (see the comments on this PR: https://github.com/OCA/server-tools/pull/922 and the docstring of the `set_groups_from_roles` method).
Now, a thing that could be done is to make this configuration through a `ir.config_parameter` here to tune the value of the `force` parameter, we keep the default behaviour if the parameter is set (or unset) then wo do not remove groups when we remove the last role(s).

What do you think?

sebalix (changes requested):

PR #114, opened 2021-03-16
 [FIX] base_user_role : affect correct default category, when creating the related group

legalsylvain:
That PR fixes the module ``base_user_role``.

The module creates a ``ir.module.category`` for groups that are generated by roles. [See](https://github.com/OCA/server-backend/blob/12.0/base_user_role/data/ir_module_category.xml#L6).

**before that PR**

However, the current default + related design doesn't work. [See](https://github.com/OCA/server-backend/blob/12.0/base_user_role/models/role.py#L36). So the groups doesn't have a category, and are displayed in the "Other" section, with other "orphan" groups.

**After that PR**

- the module.category is displayed in the user form view
- all the groups are displayed in the correct section.

image

Note 1 : ~~the patch provides a migration script to fix data on existing databases, pinned to ``12.0.3.0.0``. To maintainers : please call ocabot with ``major`` option.~~
EDIT :  the patch provides a migration script to fix data on existing databases, pinned to ``12.0.2.1.0``. To maintainers : please call ocabot with ``nobump`` option.

Note 2 : the test is written with the new design introduced by this PR #64 by @oerp-odoo. Once merged, the PR should be rebased to make travis green.

CC : @StefanRijnhart, @jcdrubay, @sebalix

### DIFF
--- a/base_user_role/models/role.py
+++ b/base_user_role/models/role.py
@@ -31,15 +31,17 @@ class ResUsersRole(models.Model):
         string="Users list",
         compute="_compute_user_ids",
     )
+    # TODO, remove in next version as it is not used in the whole module
+    # kept here for legacy reason
     group_category_id = fields.Many2one(
+        comodel_name="ir.module.category",
         related="group_id.category_id",
-        default=lambda cls: cls.env.ref(
-            "base_user_role.ir_module_category_role"
-        ).id,
         string="Associated category",
-        help="Associated group's category",
     )
     comment = fields.Html(string="Internal Notes")
+
+    def _default_category_id(self):
+        return self.env.ref("base_user_role.ir_module_category_role")
 
     @api.multi
     @api.depends("line_ids.user_id")
@@ -49,6 +51,8 @@ class ResUsersRole(models.Model):
 
     @api.model
     def create(self, vals):
+        if "category_id" not in vals and "group_id" not in vals:
+            vals.update({"category_id": self._default_category_id().id})
         new_record = super(ResUsersRole, self).create(vals)
         new_record.update_users()
         return new_record
--- a/base_user_role/__manifest__.py
+++ b/base_user_role/__manifest__.py
@@ -3,7 +3,7 @@
 
 {
     "name": "User roles",
-    "version": "12.0.2.0.1",
+    "version": "12.0.2.1.0",
     "category": "Tools",
     "author": "ABF OSIELL, Odoo Community Association (OCA)",
     "license": "AGPL-3",
--- a/base_user_role/data/ir_module_category.xml
+++ b/base_user_role/data/ir_module_category.xml
@@ -3,8 +3,9 @@
      License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
 <odoo>
 
-<record model="ir.module.category" id="ir_module_category_role">
-    <field name='name'>User roles</field>
-</record>
+    <record model="ir.module.category" id="ir_module_category_role">
+        <field name="name">User roles</field>
+        <field name="sequence">200</field>
+    </record>
 
 </odoo>
--- a/base_user_role/migrations/12.0.2.1.0/post-migration.py
+++ b/base_user_role/migrations/12.0.2.1.0/post-migration.py
@@ -1,0 +1,21 @@
+# Copyright (C) 2021 - Today: GRAP (http://www.grap.coop)
+# @author: Sylvain LE GAL (https://twitter.com/legalsylvain)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+import logging
+from odoo import api, SUPERUSER_ID
+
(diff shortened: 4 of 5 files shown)


base_user_role/models/role.py
```diff
@@ -49,6 +51,7 @@ def _compute_user_ids(self):
 
     @api.model
     def create(self, vals):
+        vals.update({"category_id": self._default_category_id().id})
```
StefanRijnhart: Please, only update `vals` if no value is present yet. In my own setup I create roles with a category already to make up for this bug.

base_user_role/migrations/12.0.3.0.0/post-migration.py
```diff
@@ -0,0 +1,21 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+import logging
+from odoo import api, SUPERUSER_ID
+
+_logger = logging.getLogger(__name__)
+
+
+def _affect_group_to_category(env):
+    _logger.info("Set Category 'User roles' to groups related to roles")
+    roles = env["res.users.role"].search([])
```
StefanRijnhart: Here too, only select roles that have no category set please.

legalsylvain: Hi @StefanRijnhart it is handled in the line 13. Or I missed something ? ``.filtered(lambda x: not x.category_id)``

base_user_role/models/role.py
StefanRijnhart: Maybe also check if the group has a category already.

base_user_role/migrations/12.0.3.0.0/post-migration.py
StefanRijnhart: Indeed! Sorry, did not look carefully.

base_user_role/models/role.py
legalsylvain: > only update vals if no value is present yet

done. Thanks.

> Maybe also check if the group has a category already.

Not sure to understand. group is created when role is created (via ``    _inherits = {"res.groups": "group_id"}``). 
Do you want I handle the case where create is called with a ``group_id`` that has a ``category_id`` defined ? 
I don't think create should be called with a vals that contains ``group_id`` key.

StefanRijnhart: On the contrary. In our module, we have
```
    <record id="group_role_something" model="res.groups">
        <field name="name">Role Someting</field>
        <field name="category_id" ref="module_category_roles"/>
        <field name="implied_ids">
        ....
        </field>
    </record>
    <record id="role_something" model="res.users.role">
        <field name="group_id" ref="module.group_role_something"/>
    </record>
```

legalsylvain: 1) OK. I had not thought of such a design. Naive question, why don't you just define roles ? (without groups, that will be created automatically)
something like : 
```
    <record id="role_something" model="res.users.role">
        <field name="name">Role Someting</field>
        <field name="line_ids">
        ....
        </field>
    </record>
```

2) I updated my code. I don't affect "category_id" if "group_id" is defined. (because I suppose that the group is correctly set, and should not be updated). Is it OK for you ?

StefanRijnhart: Thank you!

We create the group separately because we like to have explicit group xml ids so that we can use them in access rules and view definitions.

StefanRijnhart (approved):

chusamo:
LGTM Thankk you!

chusamo (approved):
LGTM Thank you!

StefanRijnhart:
Was going to merge but I saw you assigned a new major version 12.0.3.0.0 instead of a minor increment 12.0.2.2.0. I think the latter is more appropriate. Because there is a migration script with a fixed version, I would prefer it if you updated the version in the manifest manually, and then we can ask gitbot to merge using `nobump`.

dreispt:
@legalsylvain Can you remove the changes to the version, so that we can let the mergebot do that increment?

legalsylvain:
> Was going to merge but I saw you assigned a new major version 12.0.3.0.0 instead of a minor increment 12.0.2.2.0. I think the latter is more appropriate. Because there is a migration script with a fixed version, I would prefer it if you updated the version in the manifest manually, and then we can ask gitbot to merge using nobump.

You're right. reading the OCA doc, it seems more like a "minor" change.

https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#12version-numbers

> @legalsylvain Can you remove the changes to the version, so that we can let the mergebot do that increment?

No. It's safer to bump manually the ``__manifest__.py`` file and call with ``nobump``, when migration scripts are provided. So we'll have to solve conflict if another PR is merged, before this one. 

@StefanRijnhart : I bumped to ``12.0.2.1.0`` manifest and folder name. I think it's OK now.

StefanRijnhart:
Perfect, thanks!

/ocabot merge nobump

PR #116, opened 2021-03-18
[14.0] group_backend: new module

petrus-v:
Parts of the module description

This module was written to extend the standard functionality regarding users
and groups management by adding a new `Backend user` group that only gives access
to odoo backend (`/web`):

* no default access
* no default menu
* no default access rules
* no default actions
* no default views

The problem with the `Internal user` is when you want to gives access to the
backend to a really thin part of your business to some users, it's quite hard
to properly maintain those roles over the project life, a lot of models use
that group (`base.group_user`) by default which makes hard
to maintains.

So that helps creating well-defined user groups with more controls.


petrus-v:
This depends on https://github.com/odoo/odoo/issues/68106 with
odoo PR: https://github.com/odoo/odoo/pull/68111
or OCA PR: https://github.com/OCA/OCB/pull/1005

group_backend/models/res_users.py
```diff
@@ -0,0 +1,23 @@
...
+
+        This code avoid to overwrite a lot of places in controllers from
+        different modules ('portal', 'web', 'base') with hardcoded statement
+        that check if user is part of `base.group_user` group.
+
+        As far `base.group_user` have a lot of default permission this
+        makes hard to maintain proper access right according your business.
+        """
+        res = super(Users, self).has_group(group_ext_id)
+        if not res and group_ext_id == "base.group_user":
```
jcdrubay, line 25: I think this is a smart hack avoiding a lot of mess in the code.

However it makes the behavior of this module unpredictable in long term. At anytime, the native code of Odoo could be changed to rely on `has_group` to grant additional access and then we would loose control again on what the users of that group could or could not do. And since this is touching security, this is one of the few things that might change within the same major version of Odoo.

It seems that `web` module only relies one time on `self.env.user.has_group('base.group_user')` and for `base` module there are 5 cases of usage (some might not be easy to override). And it might be acceptable to not override all cases, because overriding would only make sense if you want to give the permission which might not be needed for this very special user type. So overriding only the few necessary places could make sense.

I don't think `portal` module is necessary to take into consideration as it is not really "core" (it's not auto_install anymore in v14).

In my opinion, the code of this PR could:

- stay as it is if highlighting more the risk
  OR
- add the code to override wherever necessary

To highlight the risk:

- the content of this docstring can be added as a Limitation in the Readme
- a warning log could alert that we are bypassing the expected behavior of has_group.

```
if not res and group_ext_id == "base.group_user":
    has_group_backend = super(Users, self).has_group("group_backend.group_backend")
    if has_group_backend:
        logging.waning("Forcing has_group to return True for group_backend")
    return has_group_backend
```

petrus-v: Thanks a lot for your feed backs.

I've fist had a try to override the code in different parts but was no very confident as behavior would change according module dependency tree if portal or website are installed this would make difficult to predict which one will be called last (thinking `Home._login_redirect` or `Home.index` methods) and I don't want to depends on such modules.

So I think it's very good Idea to add a warning log and prevent against risk in README limitation.

I'll do that first.

dreispt:
To follow the name conventions being used, i think the module name should be `base_group_backend`

jcdrubay (approved):

group_backend/README.rst
```diff
@@ -0,0 +1,140 @@
...
+* no default access rules
+* no default actions
+* no default views
+
+
+The problem with the `Internal user` is when you want to gives access to the
+backend to a really thin part of your business to some users, it's quite hard
+to properly maintain those roles over the project life, a lot of models use
+that group (`base.group_user`) by default which makes hard
+to maintains.
```
jcdrubay:
```suggestion
to maintain.
```

group_backend/README.rst
```diff
@@ -0,0 +1,140 @@
...
+
+So that helps creating well-defined user groups with more controls.
+
+We suggest to use this module with its compagnon `base_user_role`
+
+
+Limitations
+~~~~~~~~~~~
+
+As the time of writing Odoo use `has_group("base.group_user")` to gives the
```
jcdrubay:
```suggestion
At the time of writing, Odoo uses `has_group("base.group_user")` to give the
```

group_backend/README.rst
```diff
@@ -0,0 +1,140 @@
...
+backend access.
+This module is only overloading that method to try if user is in the
+`group_backend.group_backend` group (in case of `base.group_user`
+is requested and return `False`) to let access to the odoo backend.
+
+This avoid to overwrite a lot of overwrite in different controllers from
```
jcdrubay:
```suggestion
This avoids to write a lot of overwrite in different controllers from
```

group_backend/README.rst
```diff
@@ -0,0 +1,140 @@
...
+
+Configuration
+=============
+
+Please do not attached any rules, access, menus to that group. You should prefer
+using `implied_ids` if you really want to link it.
+
+.. note::
+
+   Be aware users can only belongs to one group from the user type category
```
jcdrubay:
```suggestion
   Be aware users can only belong to one group from the user type category
```

group_backend/README.rst
```diff
@@ -0,0 +1,140 @@
...
+   (`base.module_category_user_type`). So your group can't implied_ids to both
```
jcdrubay:
```suggestion
   (`base.module_category_user_type`). So your other groups can't inherit both
```

group_backend/README.rst
```diff
@@ -0,0 +1,140 @@
...
+   internal users and backend users. 
+
+Usage
+=====
+
+To use this module, you need to:
+
+#. Go to Configuration / Users / Users choose user and set the user type.
```
jcdrubay:
```suggestion
#. Go to Configuration / Users / Users, choose a user and set the user type.
```

jcdrubay (commented):
fix some typos

petrus-v:
> To follow the name conventions being used, i think the module name should be `base_group_backend`

Thanks for suggestion I will adding that in my todo list !

bealdav:
see 16.0 version of the module https://github.com/OCA/server-backend/pull/205

PR #119, opened 2021-04-16
 [FIX] base_user_role : role lines are global by default

legalsylvain:
Trivial patch. By default when using ``base_user_role`` in a multi company context, the expected functionality is that : 
affecting a role to a user apply the role for all the companies. (that is the odoo default functionnality). Filter per company should be an option.

This patch reset the default value.


dreispt:
Hmm, Travis is red. Maybe tests need to be adjusted?

PR #122, opened 2021-04-21
[fix] requirements.txt: Pin pymssql to 2.1.5

luisg123v:
As the more recent version 2.2.0 is not compatible with Python 3.5.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
 sqlalchemy
 radicale<3.0.0
 mysqlclient==2.0.1
-pymssql<3.0
+pymssql<=2.1.5


luisg123v:
@moylop260 could you review, please?

requirements.txt
```diff
@@ -1,4 +1,4 @@
 sqlalchemy
 radicale<3.0.0
 mysqlclient==2.0.1
-pymssql<3.0
+pymssql==2.1.5
```
moylop260: what about `<=`?

In order to avoid raising errors with new dependencies check of pip?

luisg123v: Done @moylop260

moylop260 (approved):
👍

moylop260:
@GSLabIt

What do you think?

GSLabIt (approved):

moylop260 (approved):
👍

PR #132, opened 2021-09-09
[12.0][ADD] base_default_lang_translate

remi-filament:
This module allows to select a default language (other than en_US) and. Any update performed with this language would then :

-     Update translation in this language (default behaviour)
-     Update source terms for all translations with the new value
-     Store new value in object table in database

I also have added a function to modify state of existing translations in case a source term is modified (state changes to to_translate).

This module is mainly based on https://bazaar.launchpad.net/~eoc/openobject-server/6.1-overwrite_changes_translatable_fields/revision/4377
I have added authors so I hope this is enough ?


remi-filament:
I have just pushed the module from yesterday. I ping you @moylop260 as you asked, thanks in advance for review !

moylop260 (approved):
Awesome!

Thank you so much

👍

moylop260:
We need another approve

Could you ping the other guys from thread, please?

base_default_lang_translate/models/res_lang.py
```diff
@@ -0,0 +1,16 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+
+from odoo import _, api, fields, models
+from odoo.exceptions import ValidationError
+
+
+class ResLang(models.Model):
+    _inherit = "res.lang"
+
+    default_lang = fields.Boolean("Default Language", default=False)
```
yostashiro: A non-blocking suggestion.  `is_default_lang` may be a little more explicit.

jordancoest (approved):
Thanks @remi-filament !

remi-filament:
> Could you ping the other guys from thread, please?

I could not find them on GH so I sent another e-mail on ML, thanks for approval !

base_default_lang_translate/models/res_lang.py
remi-filament: Makes sense, I have just updated it !

ypapouin (approved):
LGTM - Code Review

moylop260:
Hi @remi-filament 

If we want to use this great module for a instance already deployed we will need to generate a migrate script

Have you done this before?

remi-filament:
Hi @moylop260 ,
no I have never done that before, any help appreciated here.
If it can help I have however the script I used after installing this module for my customer (using French as default lang) but I do not know how to determine at module installation what was the default language before...

```python
# Part 1 : update terms and translations everywhere
Translations = self.env["ir.translation"]
fr_translations = Translations.search([("type", "in", ("model", "model_terms")),("lang","=","fr_FR"),("module", "in", (False, "", "__export__"))])

for fr_trans in fr_translations:
    raw_model, field = fr_trans.name.split(',')
    model = raw_model.replace('.', '_')
    if fr_trans.src != fr_trans.value and fr_trans.value != False:
        item = self.env[raw_model].with_context(lang="en_US").browse(fr_trans.res_id)
        if item.exists():
            if item[field] != fr_trans.value:
                ret_item = item.with_context(lang="en_US").write({field: fr_trans.value})
            all_lang_translations = Translations.search([('name','=',fr_trans.name),('res_id','=', fr_trans.res_id)])
            for trans in all_lang_translations:
                if trans.src != fr_trans.value:
                    trans.src = fr_trans.value
        else:
            obsolete_trans = Translations.search([('name','=',fr_trans.name),('res_id','=', fr_trans.res_id)])
            ret_unlink = obsolete_trans.unlink()

self.env.cr.commit()

# Part 2 :
# Install en_GB language (to replace en_UK to get translations in English)
# Move translations to english UK :
Translations = self.env["ir.translation"]
us_translations = Translations.search([('lang','=','en_US')])
for us_trans in us_translations:
    us_trans.lang = 'en_GB'

self.env.cr.commit()

# Part 3 : (from web UI)
# Install base_default_language_translation
# Configure fr_FR as default Language
# Deactivate en_US language
```

PR #137, opened 2021-11-03
[FIX] base_user_role_company: wrong xpath

kos94ok-3D:
<pre>
Field "active_role" does not exist in model "res.users"

View name: res.users.form.inherit.company
Error context:
 view: ir.ui.view(2849,)
 xmlid: view_res_users_form_inherit_company
 view.model: res.users
 view.parent: ir.ui.view(2752,)
 file: /home/andrew/odoo/oca14/server-backend/base_user_role_company/views/role.xml
</pre>

### DIFF
--- a/base_user_role_company/views/role.xml
+++ b/base_user_role_company/views/role.xml
@@ -5,10 +5,13 @@
         <field name="model">res.users</field>
         <field name="inherit_id" ref="base_user_role.view_res_users_form_inherit" />
         <field name="arch" type="xml">
-            <field name="role_id" position="after">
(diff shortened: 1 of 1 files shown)


kos94ok-3D:
Could you review it?
@Chandresh-OSI @hbrunn @dreispt

hbrunn (approved):

HaraldPanten (approved):

dreispt:
So we are now required to use xpath to extend inside one2many? This wasn't so before, right?

PR #141, opened 2021-12-29
12.0 mig base_dav

cesarlr:


cesarlr:
@hbrunn @fkantelberg What do you think? Ready to merge?

base_dav/models/dav_collection.py
```diff
@@ -0,0 +1,300 @@
...
+        if 'rev' not in vobj.contents and 'write_date' in record._fields:
+            vobj.add('rev').value = record.write_date.strftime('%Y-%m-%dT%H%M%SZ')
+        return result
+
+    @api.model
+    def _odoo_to_http_datetime(self, value):
+        return time.strftime(
+            '%a, %d %b %Y %H:%M:%S GMT',
+            value.timetuple(),
+        )
```
fkantelberg: value should be already a datetime object which means you can directly use `value.strftime('%a, %d %b %Y %H:%M:%S GMT')`.

fkantelberg (changes requested):
Sadly I can't test it currently. But as far as I remember the only problem was that the datetime fields changed from str to datetime objects between v11 and v12. I see that you have provided a fix there but it could be written simpler.

hbrunn (changes requested):
you'll need to increase coverage for this to be mergeable.

I'd also suggest to update to the 3.x branch of radicale: https://radicale.org/v3.html

base_dav/models/dav_collection.py
cesarlr: > you'll need to increase coverage for this to be mergeable.
> 
> I'd also suggest to update to the 3.x branch of radicale: https://radicale.org/v3.html
Radicale v3 needs Python 3.6. I have had to change Python version in Travis CI because in repo it's 3.5. I don't know if it is a good idea...

cesarlr:
> you'll need to increase coverage for this to be mergeable.
> 
> I'd also suggest to update to the 3.x branch of radicale: https://radicale.org/v3.html

Radicale v3 needs Python 3.6. I have had to change Python version in Travis CI because in repo it's 3.5. I don't know if it is a good idea...

base_dav/controllers/main.py
```diff
@@ -0,0 +1,64 @@
...
+    )
+    def handle_dav_request(self, davpath=None):
+        config = ConfigParser()
+        for section, values in radicale.config.DEFAULT_CONFIG_SCHEMA.items():
+            config.add_section(section)
+            for key, data in values.items():
+                config.set(section, key, data["value"])
+        config.set('auth', 'type', 'odoo.addons.base_dav.radicale.auth')
+        config.set(
+            'storage', 'type', 'odoo.addons.base_dav.radicale.collection'
```
oyale, line 39: It looks like we have to define the `Storage` class in [radicale/collection.py](https://github.com/OCA/server-backend/pull/141/files#diff-8e88ccea56553f5cdf667251d4f666d68e54912735688f409cf9bbd8d0e8391a) and implement the `BaseStorage` methods from `radicale.storage` if we want to change the default values.

Otherwise, we can remove this line as the L36 loop will set the default values.

I am getting a permission error while creating directories in the default path  (`/var/lib/radicale`).

base_dav/tests/test_base_dav.py
```diff
@@ -0,0 +1,119 @@
...
+from ..controllers.main import Main as Controller
+
+MODULE_PATH = "odoo.addons.base_dav"
+CONTROLLER_PATH = MODULE_PATH + ".controllers.main"
+RADICALE_PATH = MODULE_PATH + ".radicale"
+
+ADMIN_PASSWORD = "RadicalePa$$word"
+
+
+@mute_logger("radicale")
```
oyale:
```suggestion

```

base_dav/radicale/rights.py
```diff
@@ -0,0 +1,31 @@
...
+    from radicale.rights.authenticated import Rights as AuthenticatedRights
+    from radicale.rights.owner_write import Rights as OwnerWriteRights
+except ImportError:
+    AuthenticatedRights = OwnerOnlyRights = OwnerWriteRights = None
+
+
+class Rights(OwnerOnlyRights, OwnerWriteRights, AuthenticatedRights):
+    def authorization(self, user, path):
+        if path == '/':
+            return True
```
oyale, line 17:
```suggestion
            return ""
```

base_dav/radicale/rights.py
```diff
@@ -0,0 +1,31 @@
...
+
+        collection = Collection(path)
+        if not collection.collection:
+            return False
```
oyale:
```suggestion
            return ""
```

base_dav/radicale/rights.py
```diff
@@ -0,0 +1,31 @@
...
+
+        rights = collection.collection.sudo().rights
+        cls = {
+            "owner_only": OwnerOnlyRights,
+            "owner_write_only": OwnerWriteRights,
+            "authenticated": AuthenticatedRights,
+        }.get(rights)
+        if not cls:
+            return False
```
oyale, line 30:
```suggestion
            return ""
```

base_dav/radicale/rights.py
oyale: This method must return granted rights (e.g. ``"RW"``), not a boolean.

oyale (changes requested):

base_dav/controllers/main.py
```diff
@@ -0,0 +1,64 @@
...
+    @http.route(
+        [PREFIX, '%s/<path:davpath>' % PREFIX], type='http', auth='none',
+        csrf=False,
+    )
+    def handle_dav_request(self, davpath=None):
+        config = ConfigParser()
+        for section, values in radicale.config.DEFAULT_CONFIG_SCHEMA.items():
+            config.add_section(section)
+            for key, data in values.items():
+                config.set(section, key, data["value"])
```
oyale:
```suggestion
                config.set(section, key, data["value"] if type(data) == dict else data)
```

base_dav/controllers/main.py
```diff
@@ -0,0 +1,64 @@
...
+        config.set('auth', 'type', 'odoo.addons.base_dav.radicale.auth')
+        config.set(
+            'storage', 'type', 'odoo.addons.base_dav.radicale.collection'
+        )
+        config.set(
+            'rights', 'type', 'odoo.addons.base_dav.radicale.rights'
+        )
+        config.set('web', 'type', 'none')
+        application = radicale.Application(
+            config, logging.getLogger('radicale'),
```
oyale:
```suggestion
            config
```

oyale (changes requested):
This will cause an error (3 instead 2 positional arguments)

diggy128:
Anything I can do to help speed this up?
Merp from Xpansa fails to sync contacts and calendar on Android and it is essential in for implementation.

oyale:
> Anything I can do to help speed this up? Merp from Xpansa fails to sync contacts and calendar on Android and it is essential in for implementation.

Hi @diggy128! Thanks for asking.

We need to improve code coverage to migrate the add-on into v12. Could you help us with that?

diggy128:
Hi @oyale
Writing tests is not my strong point and this time of the year I'm very time stressed but I could find some time eventually to chip in.

On the other hand, I downloaded the PR and I have come up with a lot of problems. It even seems that the module doesn't work at all.

Maybe we should look on that first or have I done something completely wrong that broke my installation?
I'll put some comments on the code to give you an insight.

oyale:
> Hi @oyale Writing tests is not my strong point and this time of the year I'm very time stressed but I could find some time eventually to chip in.
> Maybe we should look on that first or have I done something completely wrong that broke my installation? I'll put some comments on the code to give you an insight.

You're absolutelly right. 

Sadly, we've run out of time for this issue this term. Any insights would be greatly appreciated.

huguesdk:
i’ve done the porting to 12.0 independently in #226 and discovered this pr only after that. in that pr i’ve only ported the code to work with odoo 12, without upgrading to radicale 3. i propose to make the module already available in 12 in its current state, so that it can already be used. i think that the port to radicale 3 should be handled in a separate subsequent pr, as it is not directly linked to the porting to odoo 12. what do you think?

oyale:
I agree that handling the migration to radicale 3 in a separate PR would allow to use the module now and also, it is not a requirement to port it to odoo 12.

diggy128:
Same thing here. Having a module that works, even with a bit of outdated dependency is better than no module at all.

huguesdk:
thanks for your answers. feel free to review #226 so that it can be merged quickly.

PR #142, opened 2022-01-18
[13.0][FIX] base_import_match - fix importing one2many records

Rad0van:
When trying to use `base_import_match` and importing related one2many records it fails badly with not really helpful message. Let's have a look at this code here:
https://github.com/OCA/server-backend/blob/0414dd0b0530389862a3da142b5642641c0cabfb/base_import_match/models/base.py#L10-L31

Imagine we're importing `product.template` and have a rule setup to match `default_code`. We want to update `price` and also create new `product.supplierinfo` record - it is one2many attribute named named `seller_ids`. When mapped through import GUI something like this comes in:

```
fields = ['default_code', 'price', 'seller_ids/name/.id', 'seller_ids/min_qty', 'seller_ids/price', 'seller_ids/date_start']
data = [['TEST', '755', '11', '1', '566.25', '2022-01-18']]
```
That is all correct. Next comes these two lines:
https://github.com/OCA/server-backend/blob/0414dd0b0530389862a3da142b5642641c0cabfb/base_import_match/models/base.py#L22-L25
Result of that is something like this:
```
import_fields = [['default_code'], ['price'], ['seller_ids', 'name', '.id'], ['seller_ids', 'min_qty'], ['seller_ids', 'price'], ['seller_ids', 'date_start']]

converted_data = 
[(False,
  False,
  {'default_code': 'TEST',
   'price': 755.0,
   'seller_ids': [(0,
                   False,
                   {'date_start': '2022-01-18',
                    'min_qty': 1.0,
                    'name': 11,
                    'price': 566.25})]},
  {'record': 0, 'rows': {'from': 0, 'to': 0}})]
```
So far so good. But then comes the part that ruins all:
https://github.com/OCA/server-backend/blob/0414dd0b0530389862a3da142b5642641c0cabfb/base_import_match/models/base.py#L31
The result of that is something like this:
`clean_fields = ['default_code', 'price', 'seller_ids', 'seller_ids', 'seller_ids', 'seller_ids']`
The names of attributes of related record are all lost. This later leads to an attempt to map the same value (last of the related attribute?) to all related attributes:
https://github.com/OCA/server-backend/blob/0414dd0b0530389862a3da142b5642641c0cabfb/base_import_match/models/base.py#L51

So this is an attempt to fix this. I have tested this through GUI as well as created a test case (not very pretty) that fails on original code.

@sbidoul @Yajo @pedrobaeza @sergio-teruel @zaoral @chienandalu please have a look. And have mercy on me please :-)

### DIFF
--- a/base_import_match/models/base.py
+++ b/base_import_match/models/base.py
@@ -26,29 +26,33 @@ class Base(models.AbstractModel):
             # Mock Odoo to believe the user is importing the ID field
             if "id" not in fields:
                 fields.append("id")
-                import_fields.append(["id"])
             # Needed to match with converted data field names
-            clean_fields = [f[0] for f in import_fields]
             for dbid, xmlid, record, info in converted_data:
-                row = dict(zip(clean_fields, data[info["record"]]))
-                match = self
-                if xmlid:
-                    # Skip rows with ID, they do not need all this
-                    row["id"] = xmlid
-                    continue
-                elif dbid:
-                    # Find the xmlid for this dbid
-                    match = self.browse(dbid)
-                else:
-                    # Store records that match a combination
-                    match = self.env["base_import.match"]._match_find(self, record, row)
-                # Give a valid XMLID to this row if a match was found
-                # To generate externals IDS.
-                match.export_data(fields)
-                ext_id = match.get_external_id()
-                row["id"] = ext_id[match.id] if match else row.get("id", u"")
-                # Store the modified row, in the same order as fields
-                newdata.append(tuple(row[f] for f in clean_fields))
+                # In case of one2many on empty lines one record may contain several rows
+                for row_index in range(info["rows"]["from"], info["rows"]["to"] + 1):
+                    row = dict(zip(fields, data[row_index]))
+                    match = self
+                    if xmlid:
+                        # Skip rows with ID, they do not need all this
+                        row["id"] = xmlid
+                        continue
+                    elif dbid:
+                        # Find the xmlid for this dbid
+                        match = self.browse(dbid)
+                    elif row_index == info["rows"]["from"]:
+                        # Store records that match a combination
+                        # But only for first row of record,
+                        # because the rest contain one2many fields
+                        match = self.env["base_import.match"]._match_find(
+                            self, record, row
+                        )
+                    # Give a valid XMLID to this row if a match was found
+                    # To generate externals IDS.
+                    match.export_data(fields)
+                    ext_id = match.get_external_id()
+                    row["id"] = ext_id[match.id] if match else row.get("id", u"")
+                    # Store the modified row, in the same order as fields
+                    newdata.append(tuple(row[f] for f in fields))
             # We will import the patched data to get updates on matches
             data = newdata
         # Normal method handles the rest of the job
--- a/base_import_match/readme/CONTRIBUTORS.rst
+++ b/base_import_match/readme/CONTRIBUTORS.rst
@@ -2,3 +2,4 @@
   * Jairo Llopis
   * Vicent Cubells
   * Ernesto Tejeda
+* Radovan Skolnik <radovan@skolnik.info>
--- a/base_import_match/tests/import_data/res_partner_email_one2many.csv
+++ b/base_import_match/tests/import_data/res_partner_email_one2many.csv
@@ -1,0 +1,4 @@
+email,function,child_ids/name,child_ids/color,child_ids/email
+floyd.steward34@example.com,Bug Fixer,Bart Steward,666,bart.steward@example.com
+,,Lisa Steward,777,lisa.steward@example.com
+,,Maggie Steward,555,maggie.steward@example.com
--- a/base_import_match/tests/test_import.py
+++ b/base_import_match/tests/test_import.py
@@ -68,3 +68,41 @@ class ImportCase(TransactionCase):
         record = self._base_import_record("res.users", "res_users_login")
         record.do(["login", "name"], [], OPTIONS)
         self.assertEqual(self.env.ref("base.user_demo").name, "Demo User Changed")
+
+    def test_res_partner_email_one2many(self):
+        """Change function based on email and import one2many record."""
+        record = self._base_import_record("res.partner", "res_partner_email_one2many")
+        record.do(
+            [
+                "email",
+                "function",
+                "child_ids/name",
+                "child_ids/color",
+                "child_ids/email",
+            ],
+            [],
+            OPTIONS,
+        )
+        self.assertEqual(
+            self.env.ref("base.res_partner_address_4").function, "Bug Fixer"
+        )
+        self.assertTrue(self.env.ref("base.res_partner_address_4").child_ids,)
+        self.assertEqual(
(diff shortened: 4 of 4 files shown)


Rad0van:
I wonder how that (double post) happened. Must have been the late hour...

pedrobaeza (approved):

pedrobaeza:
Please use proper commit message: https://www.odoo.com/documentation/15.0/es/developer/misc/other/guidelines.html#commit-message-structure

Rad0van:
> Please use proper commit message: https://www.odoo.com/documentation/15.0/es/developer/misc/other/guidelines.html#commit-message-structure

Done. It was just few days you pointed me to that and I already forgot. But was quite late. Hope it's OK now.

Rad0van:
@pedrobaeza while this makes it better it is far from perfect. It only works when there is only one such record to be imported. If there is more it crashes like this:
2022-01-20 22:19:39,601 16106 INFO odoo13com odoo.addons.base_import.models.base_import: importing 3 rows... 
2022-01-20 22:19:41,741 16106 ERROR odoo13com odoo.http: Exception during JSON request handling. 
```
  File "/home/rex/Odoo/odoo-13.0/odoo/http.py", line 624, in _handle_exception
    return super(JsonRequest, self)._handle_exception(exception)
  File "/home/rex/Odoo/odoo-13.0/odoo/http.py", line 310, in _handle_exception
    raise pycompat.reraise(type(exception), exception, sys.exc_info()[2])
  File "/home/rex/Odoo/odoo-13.0/odoo/tools/pycompat.py", line 14, in reraise
    raise value
  File "/home/rex/Odoo/odoo-13.0/odoo/http.py", line 669, in dispatch
    result = self._call_function(**self.params)
  File "/home/rex/Odoo/odoo-13.0/odoo/http.py", line 350, in _call_function
    return checked_call(self.db, *args, **kwargs)
  File "/home/rex/Odoo/odoo-13.0/odoo/service/model.py", line 94, in wrapper
    return f(dbname, *args, **kwargs)
  File "/home/rex/Odoo/odoo-13.0/odoo/http.py", line 339, in checked_call
    result = self.endpoint(*a, **kw)
  File "/home/rex/Odoo/odoo-13.0/odoo/http.py", line 915, in __call__
    return self.method(*args, **kw)
  File "/home/rex/Odoo/odoo-13.0/odoo/http.py", line 515, in response_wrap
    response = f(*args, **kw)
  File "/mnt/slow/rex/Programming/odoo/odoo-13.0/addons/web/controllers/main.py", line 1342, in call_kw
    return self._call_kw(model, method, args, kwargs)
  File "/mnt/slow/rex/Programming/odoo/odoo-13.0/addons/web/controllers/main.py", line 1334, in _call_kw
    return call_kw(request.env[model], method, args, kwargs)
  File "/home/rex/Odoo/odoo-13.0/odoo/api.py", line 390, in call_kw
    result = _call_kw_multi(method, model, args, kwargs)
  File "/home/rex/Odoo/odoo-13.0/odoo/api.py", line 377, in _call_kw_multi
    result = method(recs, *args, **kwargs)
  File "/mnt/slow/rex/Programming/odoo/odoo-13.0/addons/base_import/models/base_import.py", line 924, in do
    import_result = model.load(import_fields, data)
  File "/mnt/slow/rex/Programming/odoo/odoo-13.0/.addons-local/base_import_match/models/base.py", line 33, in load
    for dbid, xmlid, record, info in converted_data:
  File "/home/rex/Odoo/odoo-13.0/odoo/models.py", line 1180, in _convert_records
    for record, extras in stream:
  File "/home/rex/Odoo/odoo-13.0/odoo/tools/misc.py", line 883, in next
    val = next(self.stream, _ph)
  File "/home/rex/Odoo/odoo-13.0/odoo/models.py", line 1129, in _extract_records
    record_span = list(itertools.chain([row], record_span))
  File "/home/rex/Odoo/odoo-13.0/odoo/models.py", line 1113, in only_o2m_values
    return any(get_o2m_values(row)) and not any(get_nono2m_values(row))
IndexError: list index out of range
```

I am trying to wrap my head around it for many hours now but so far nothing. I am getting pretty desperate...

Rad0van:
@pedrobaeza I have found this only works partially. For one record it works fine. But if there are more records, it fails. After few hours of debugging I found that the culprit lies in cleverness of import that allows to import multiple one2many records on lines following the first one with other fields blank. The code that attempts to detect such situation is part of `_extract_records` that is being called before processing each record in our code:
https://github.com/odoo/odoo/blob/7504b78625f79bb7cf69f8c5a0502ac924fd9cab/odoo/models.py#L1124-L1127
And here's definition of that `only_o2m_values`:
https://github.com/odoo/odoo/blob/7504b78625f79bb7cf69f8c5a0502ac924fd9cab/odoo/models.py#L1111-L1113
As there already is `id` added to the field list but not the value itself for it, that `get_nono2m_values(row)` ends up with `IndexError: list index out of range`.

Now I am scratching my head and not having a clue how to overcome this. As it fails in core Odoo code we should probably alter the data we send there. So we should pre-process them in some way taking that core code into consideration. Any idea?

Rad0van:
The last commit should fix that. Hopefully. I will create some more test cases to be sure.

Rad0van:
Still does not work reliably. So far it seems if I have a "main" line followed by let's say 3 lines of related one2many fields, only 2 of them get created. Will investigate more.

Rad0van:
OK, this should be it. Now the import should behave exactly like a normal one including possibility to import one2many related fields on empty lines. Because when importing such record the `data` that enter the function look like this:

```
[
 ['floyd.steward34@example.com',
  'Bug Fixer',
  'Floyd Stewart Jr.',
  '666',
  'floyd.steward.jr@example.com',
  '7'],
 ['', '', 'Brenda Stewart', '777', 'brenda.stewart@example.com', '7'],
 ['', '', 'Homer Stewart', '555', 'homer.stewart@example.com', '7']
]
```
What we receive from `converted_data = self._convert_records(self._extract_records(import_fields, data))` looks like this:
```
[(False,
  False,
  {'child_ids': [(0,
                  False,
                  {'color': 666,
                   'email': 'floyd.steward.jr@example.com',
                   'name': 'Floyd Stewart Jr.',
                   'user_id': 7}),
                 (0,
                  False,
                  {'color': 777,
                   'email': 'brenda.stewart@example.com',
                   'name': 'Brenda Stewart',
                   'user_id': 7}),
                 (0,
                  False,
                  {'color': 555,
                   'email': 'homer.stewart@example.com',
                   'name': 'Homer Stewart',
                   'user_id': 7})],
   'email': 'floyd.steward34@example.com',
   'function': 'Bug Fixer'},
  {'record': 0, 'rows': {'from': 0, 'to': 2}})]
```
The `info` at the end tells us it is record 0 but it actually contains 3 rows - 0 to 2. So we need to process them all but only do the magic with id in the first one. The rest is one2many fields where the id cannot be used otherwise it would overwrite the parent record with empty values (in this case email and function of Floyd Steward).

pedrobaeza (approved):

Rad0van:
This should be fine to go and improves this module greatly. @sbidoul @Yajo @sergio-teruel @zaoral @chienandalu would you please care to have a look? Thank you.

PR #146, opened 2022-01-31
[14.0][FIX] base_user_role: text on role Internal Notes was not saved

dreispt:
Editing the `comment`field on the form dis not work.
It  seems that this is because the field already exists in `res.groups` with the same name, and was being shadowed.

### DIFF
--- a/base_user_role/models/role.py
+++ b/base_user_role/models/role.py
@@ -33,9 +33,6 @@ class ResUsersRole(models.Model):
         help="Associated group's category",
         readonly=False,
     )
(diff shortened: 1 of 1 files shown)


sebalix (approved):
Indeed: https://github.com/odoo/odoo/blob/2fe3920207e03fc483230d78b842320a837b5d7d/odoo/addons/base/models/res_users.py#L150
We are losing the HTML feature but better to not duplicate this field and I suppose that it never worked on 14.0... so :+1:

PR #151, opened 2022-03-09
[WIP][15][ADD] base_external_api

Vimal-OSI:


base_external_api/models/apicli_message.py
```diff
@@ -0,0 +1,22 @@
+# Copyright (C) 2022 Open Source Integrators
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+from odoo import fields, models
+
+
+class ApicliMessage(models.Model):
+    _name = "apicli.message"
+    _description = "API Client message"
+    _rec_name = "connection_id"
```
dreispt: endpoint is a better choice

base_external_api_ftp/demo/api_connection_ftp_demo.xml
```diff
@@ -9,4 +9,17 @@
         <field name="password">admin</field>
     </record>
 
+    <record id="delete_ftp_file" model="ir.cron">
+        <field name="name">Delete FTP File</field>
+        <field name="model_id" ref="base_external_api_ftp.model_apicli_connection" />
+        <field name="state">code</field>
+        <field
+            name="code"
+        >model.cron_delete_ftp_file(subdirectory='/tmp', connection_id=env.ref('base_external_api_ftp.api_connection_ftp_demo'))</field>
```
dreispt: Same comment as I did on a separate PR: use the connrction as the object operated on, instead  of an argument.

base_external_api/models/apicli_message.py
```diff
@@ -16,6 +16,7 @@ class ApicliMessage(models.Model):
             ("draft", "Draft"),
             ("done", "Done"),
             ("error", "Error"),
+            ("todo", "TODO"),
```
dreispt: The sequence is to have "todo"  just after draft

base_external_api_ftp/models/apicli_connection_ftp.py
```diff
@@ -18,27 +18,36 @@ class ApicliConnection(models.Model):
...
+    def create_delete_file_message(self, ftp_session, subdirectory):
         from_path, subdir_list, file_list = next(os.walk(subdirectory or "/tmp"))
         for file_name in file_list:
             file_path = os.path.join(from_path, file_name)
-            state = "draft"
+            message_id = self.env["apicli.message"].create(
+                {
+                    "connection_id": self.id,
+                    "endpoint": file_name,
+                    "content": "",
```
dreispt: We are missing reading the file content

base_external_api_ftp/models/apicli_connection_ftp.py
```diff
@@ -18,27 +18,36 @@ class ApicliConnection(models.Model):
...
+                    "state": "draft",
+                }
+            )
             try:
                 delete_msg = ftp_session.delete(file_name)
-                state = "done"
+                message_id.write(
+                    {
+                        "content": delete_msg,
```
dreispt: delete_msg is not the file content, it is the delete action response. So we're missing the file content.

base_external_api_ftp/models/apicli_connection_ftp.py
```diff
@@ -18,27 +18,36 @@ class ApicliConnection(models.Model):
...
+                {
+                    "connection_id": self.id,
+                    "endpoint": file_name,
+                    "content": "",
+                    "state": "draft",
+                }
+            )
             try:
                 delete_msg = ftp_session.delete(file_name)
-                state = "done"
```
dreispt: If delete is succesful, set as "todo". 
That means that it is ready for the future message processing method to process it.

base_external_api_ftp/models/apicli_connection_ftp.py
```diff
@@ -18,27 +18,36 @@ class ApicliConnection(models.Model):
...
+                        "state": "done",
+                    }
+                )
             except ftplib.error_perm as error:
                 delete_msg = error
-                state = "error"
+                message_id.write(
+                    {
+                        "content": delete_msg,
+                        "state": "todo",
```
dreispt: If it errored, I would set it as cancelled instead.

base_external_api_ftp/models/apicli_connection_ftp.py
```diff
@@ -18,27 +18,36 @@ class ApicliConnection(models.Model):
         selection_add=[("ftp", "FTP")], ondelete={"ftp": "set default"}
     )
 
-    def delete_ftp_file(self, ftp_session, subdirectory):
+    def create_delete_file_message(self, ftp_session, subdirectory):
```
dreispt: Current structure will have performance issues.
As a first step to address that, please split the create and delete logic in two separate methods.
Thanks

base_external_api_ftp/models/apicli_connection_ftp.py
```diff
@@ -75,7 +77,8 @@ def _download_each_file(self, subdirectory, ftp):
 
     @api.model
     def cron_download_ftp_files(self, subdirectory="/", conn_code=None):
-        conn_code = "DemoFTP"
+        if conn_code is None:
+            conn_code = "DemoFTP"
```
dreispt: @diogocsc  Instead Do not hard code this. Instead add a demo data file, that writes on the default cron record, so that it calls `cron_download_ftp_files(conn_code="DemoFTP")`

base_external_api/views/res_partner_view.xml
```diff
@@ -20,4 +20,13 @@
         <field name="code">records.action_upload_partner_master()</field>
     </record>
 
+    <record id="action_download_ftp" model="ir.actions.server">
+        <field name="name">FTP Download Demo</field>
+        <field name="model_id" ref="base.model_res_partner" />
+        <field name="binding_model_id" ref="base.model_res_partner" />
+        <field name='groups_id' eval="[(4, ref('base.group_no_one'))]" />
+        <field name="state">code</field>
+        <field name="code">records.action_download_partner_master()</field>
```
dreispt: Maybe this should be in a demo data file.
In that case you could `records.action_download_partner_master(conn_code="DemoFTP")`

dreispt: Better yet, "records.action_download_partner_master(subdirectory="/Inbox", conn_code="DemoFTP")

base_external_api_ftp/models/apicli_connection_ftp.py
```diff
@@ -112,7 +112,7 @@ def _ftp_upload_directory(self, ftp_session, from_local_dir, to_server_dir):
                 ftp_session, os.path.join(from_local_dir, subdir), subdir
             )
 
-    def _send_ftp_upload(self, from_local_dir, to_server_dir="."):
+    def _send_ftp_upload(self, from_local_dir, to_server_dir="Inbox"):
```
dreispt: The "Inbox" is really a configuration, and should be passed to the cron method, rather than being set as a default in code.

base_external_api_ftp/data/api_connection_ftp_data.xml
```diff
@@ -1,20 +1,15 @@
-<odoo>
+<odoo noupdate="1">
+
+    <record id="delete_ftp_file" model="ir.cron">
+        <field name="name">Delete FTP File</field>
```
dreispt: "Download" not "Delete"

base_external_api_ftp/tests/test_connection_ftp.py
```diff
@@ -0,0 +1,22 @@
...
+        "Connection is confirmed"
+        self.assertEqual(self.demo_connection.state, "confirmed")
+
+    def test_upload_file_ftp(self):
+        "tests the file upload"
+        self.assertTrue(self.partner1.action_upload_partner_master)
+
+    def test_delete_file_ftp(self):
+        "tests the files deletion"
+        self.assertTrue(self.demo_connection.cron_delete_ftp_file)
```
Nikul-OSI: **`cron_download_ftp_files`** insert of **`cron_delete_ftp_file`**

base_external_api_ftp/models/apicli_connection_ftp.py
```diff
@@ -86,7 +86,8 @@ def cron_download_ftp_files(self, subdirectory="/", conn_code=None):
                 ) as sftp:
                     conn._download_each_file(subdirectory, sftp)
 
-        self.env["apicli.message"].process_messages()
+        messages = self.env["apicli.message"].search([("state", "=", "todo")])
```
Nikul-OSI: @dreispt we already filtered message in **process_messages**.

see here: https://github.com/OCA/server-backend/pull/151/files#diff-fc037587474562c4b7f5bad821396cd3b4908f76a2d1a9e8a9c6540d2a16c179R46

dreispt: The process_messages could also be called directly. In fact, could you add that ability to the “Action” menu of the Stock Picking please?

PR #154, opened 2022-03-17
[MIG] [14.0] server_action_navigate

AshishHirapara:
Module migration to 14.0


francesco-ooops (approved):
Functional review OK

elvise (approved):
Go!

francesco-ooops:
@OCA/tools-maintainers is this good for merge?

francesco-ooops:
> @OCA/tools-maintainers is this good for merge?

@simahawk @dreispt @pedrobaeza please? :)

server_action_navigate/__manifest__.py
```diff
@@ -1,17 +1,18 @@
...
 # Copyright (C) 2020 - Today: GRAP (http://www.grap.coop)
 # @author: Sylvain LE GAL (https://twitter.com/legalsylvain)
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 {
     "name": "Server Actions - Navigate",
-    "version": "13.0.1.0.0",
-    "author": "GRAP, " "Odoo Community Association (OCA)",
+    "version": "14.0.1.0.0",
+    "author": "Ashish Hirpara, GRAP, Odoo Community Association (OCA)",
```
simahawk: your name shouldn't be here judging from the changes, listing yourself in the contributors is enough

server_action_navigate/__manifest__.py
```diff
@@ -1,17 +1,18 @@
...
     "summary": "Navigate between any items of any Odoo Models",
     "category": "Tools",
     "website": "https://github.com/OCA/server-backend",
     "license": "AGPL-3",
     "depends": ["base"],
-    "maintainers": ["legalsylvain"],
+    "maintainers": ["ashishhirpara"],
```
simahawk: @legalsylvain are we sure about this?

server_action_navigate/models/base.py
```diff
@@ -0,0 +1,62 @@
+# Migrated to v14.0 by Ashish Hirpara (https://www.ashish-hirpara.com)
```
simahawk: please remove this line, it's not your copyright here

server_action_navigate/__manifest__.py
```diff
@@ -1,17 +1,18 @@
+# Migration 2022 - 14.0: Ashish Hirpara (https://ashish-hirpara.com)
```
simahawk: remove this line pls

server_action_navigate/models/base.py
simahawk: also, I fail to find where this is coming from... any pointer?

server_action_navigate/readme/CONTRIBUTORS.rst
```diff
@@ -1 +1,2 @@
+* Ashish Hirpara (https://ashish-hirpara.com)
```
simahawk: your name after pls

simahawk (changes requested):
thanks for your contrib, please respect original authors attribution

simahawk:
Seems this commit https://github.com/OCA/server-backend/pull/154/commits/0849f2129827638b305c7ca07e1a746755ce12f3 does not belong to this work

server_action_navigate/__manifest__.py
legalsylvain: @AshishHirapara. This is a list, if you want to maintain this module no problem. (I think it's the first time modules are comaintained, but let's try !)

could you replace by ``"maintainers": ["legalsylvain", "ashishhirpara"]``

regards.

server_action_navigate/models/ir_actions_server.py
```diff
@@ -9,7 +9,9 @@
 class IrActionsServer(models.Model):
     _inherit = "ir.actions.server"
 
-    state = fields.Selection(selection_add=[("navigate", "Navigate")])
+    state = fields.Selection(
+        selection_add=[("navigate", "Navigate")], ondelete={"navigate": "set default"}
```
legalsylvain, line 13: I never saw ondelete on Selection fields. What is the interest, changing value when the module is uninstalled ?

server_action_navigate/models/ir_actions_server_navigate_line.py
```diff
@@ -22,7 +22,10 @@ class IrActionsServerNavigateLine(models.Model):
     )
 
     field_id = fields.Many2one(
-        comodel_name="ir.model.fields", string="Field", required=True
+        comodel_name="ir.model.fields",
+        string="Field",
+        required=True,
+        ondelete="cascade",
```
legalsylvain, line 28: I think this ondelete will create unconsistencies state and so doesn't see the interest...

server_action_navigate/models/base.py
legalsylvain: this function was in odoo core and was removed after 13.0 see : 

- https://github.com/odoo/odoo/blob/13.0/odoo/models.py#L4892

However, I don't know what's the best design, but copying the whole function in this module doesn't seems to be a good idea.
- maybe a dedicated module
- or remove the call, and write a work around.

@AshishHirapara, what do you think ? Do you need some help on that topic ? 

regards.

legalsylvain (changes requested):
some remarks inline.

regards.

server_action_navigate/models/base.py
AshishHirapara: > this function was in odoo core and was removed after 13.0 see :
> 
> * [https://github.com/odoo/odoo/blob/13.0/odoo/models.py#L4892](https://github.com/odoo/odoo/blob/13.0/odoo/models.py?rgh-link-date=2022-04-20T08%3A56%3A42Z#L4892)
> 
> However, I don't know what's the best design, but copying the whole function in this module doesn't seems to be a good idea.
> 
> * maybe a dedicated module
> * or remove the call, and write a work around.
> 
> @AshishHirapara, what do you think ? Do you need some help on that topic ?
> 
> regards.


Thanks, @legalsylvain. I got that the function is deleted from 14, and if you think it's a bad idea, then help me here.
You have suggested 2 solutions here.
1. A dedicated module: what do you mean by a dedicated module? Just for the function? Because it will be ultimately the same, right?
2. Remove the call: I can remove the function call and write the whole function code at the location. Will that be a proper solution?

Thanks

server_action_navigate/models/ir_actions_server.py
AshishHirapara: > 

Here I have added a new selection value to an existing field. So if the module gets uninstall, then it sets the default value, so that for the records that don't have the new value, get the default value set instead of the blank value.

server_action_navigate/models/ir_actions_server_navigate_line.py
AshishHirapara: > I think this ondelete will create unconsistencies state and so doesn't see the interest...

Odoo is not allowing this field without ondelete cascade, and shouldn't be. Because when this field gets deleted, the navigate line also should get deleted. Navigate line has no meaning without that field, just like the sale order line has no meaning without the sale order.

server_action_navigate/models/ir_actions_server.py
legalsylvain: but it doesn't make sense. to have a "navigate" ir.actions.server with a state != "navigate". 
If we uninstall the module, a better solution is to delete the ``ir.actions.server`` row where ``state == 'navigate'``.
That can be done in an uninstall_hook.

don't you think ?

server_action_navigate/__manifest__.py
AshishHirapara: Done. And I will take care of this from now on.

server_action_navigate/__manifest__.py
AshishHirapara: Sure, done. thanks.

server_action_navigate/__manifest__.py
AshishHirapara: done

server_action_navigate/readme/CONTRIBUTORS.rst
AshishHirapara: done

server_action_navigate/models/ir_actions_server.py
AshishHirapara: I have updated the change you suggested to remove the rows with uninstall hook in the last commit, but odoo is refusing the module without the ondelete param, you can see the error log here [[https://postimg.cc/svcHyDNs](https://postimg.cc/svcHyDNs)] while installing the module. So I haven't removed the ondelete parameter as of now.

AshishHirapara: @legalsylvain what do you think regarding this update?

server_action_navigate/models/ir_actions_server_navigate_line.py
legalsylvain: Agree with you. It wasn't my point.
consider a server action navigate of (sale.order) with 4 lines : 
- line 1, field 1 (line_ids), model 1 (sale.order.line)
- line 2, field 2 (product_id), model 2 (product.product)
- line 3, field 3 (product_tmpl_id) model 3 (product.template)
- line 4, field 4 (categ_id), model 4 (product.category)

If something remove the field 2 & 3, it will remove the line 2 & 3 . so the final configuration will be :
- line 1, field 1 (line_ids), model 1 (sale.order.line)
- line 4, field 4 (categ_id), model 4 (product.category)

that will be inconsistent, because ``categ_id`` is not a field of ``sale.order.line``.

That being said, I suspect that it is a theoritical case, because if field 3 is uninstalled, field 4 should also uninstalled, by dependency.

So, sorry for the noice on that topic. :+1:

server_action_navigate/models/ir_actions_server.py
legalsylvain: LGTM, Thanks !

server_action_navigate/__manifest__.py
```diff
@@ -4,14 +4,15 @@
 
 {
     "name": "Server Actions - Navigate",
-    "version": "13.0.1.0.0",
-    "author": "GRAP, " "Odoo Community Association (OCA)",
+    "version": "14.0.1.0.0",
+    "author": "Ooops, Ashish Hirpara, GRAP, Odoo Community Association (OCA)",
```
legalsylvain: nitpick : why 2 new authors, for 1 commit ?

legalsylvain (approved):
Thanks for the changes ! :+1: 
Two remaining points : 
- I guess if you removed the ``resolve_2many_commands`` call, you could remove the ``base.py`` file ?
- Changes are done on ``base_user_role_company`` module. I think there is some error in the cherry-pick process. Don't you think ?

regards, and thanks for porting this module.

server_action_navigate/models/ir_actions_server_navigate_line.py
AshishHirapara: thanks, and I got the problem, and yes it is a theoritical case and may happen rarely. So what do you suggest? Should I add the fix for it?

server_action_navigate/__manifest__.py
AshishHirapara: got it, updated. Thanks.

AshishHirapara:
> Thanks for the changes ! +1 Two remaining points :
> 
> * I guess if you removed the `resolve_2many_commands` call, you could remove the `base.py` file ?
> * Changes are done on `base_user_role_company` module. I think there is some error in the cherry-pick process. Don't you think ?
> 
> regards, and thanks for porting this module.

Thanks for pointing it out, and I have removed the `base.py` file.
But I did not get your 2nd point. Can you describe it more?

legalsylvain:
Please check thé modified files of your pr. 
https://github.com/OCA/server-backend/pull/154/files

Some are not related to the current work i think

AshishHirapara:
> Please check thé modified files of your pr. [#154 (files)](https://github.com/OCA/server-backend/pull/154/files)
> 
> Some are not related to the current work i think

Got it and fixed it.
Thanks.

legalsylvain:
Thanks @AshishHirapara could you also take a look on the conflict ?

image

AshishHirapara:
> Thanks @AshishHirapara could you also take a look on the conflict ?
> 
> image

I think I messed up something while resolving the conflict. But I am not sure, can you review it? Thanks.

server_action_navigate/__init__.py
```diff
@@ -0,0 +1,8 @@
+from . import models
```
simahawk, line 1: please move this too `hooks.py` and leave here only `from .hooks import uninstall_hook`

simahawk:
> > Thanks @AshishHirapara could you also take a look on the conflict ?
> > image
> 
> I think I messed up something while resolving the conflict. But I am not sure, can you review it? Thanks.

LG

AshishHirapara:
> thanks for your contrib, please respect original authors attribution

I will take care next time. Is any change required to do in manifest about it for now?

server_action_navigate/__init__.py
AshishHirapara: done

AshishHirapara: @simahawk moved everything from init.py to hooks.py but 'from . import models'
Because pre-commit removes the above line from hooks.py. So I kept it in init.py. Is there any other way?

simahawk: bad phrasing, sorry: import models should stay here, of course :)

```
from . import models
from .hooks import uninstall_hook
```

AshishHirapara: ok, then it's done

server_action_navigate/__manifest__.py
```diff
@@ -4,14 +4,15 @@
 
 {
     "name": "Server Actions - Navigate",
-    "version": "13.0.1.0.0",
-    "author": "GRAP, " "Odoo Community Association (OCA)",
+    "version": "14.0.1.0.0",
+    "author": "GRAP, Ooops, Odoo Community Association (OCA)",
```
simahawk: One last thing. Bare migrations do not entitle for being recognized as "authors". Being in "contributors" is enough for now.

simahawk (approved):
@legalsylvain fine for you?

server_action_navigate/__manifest__.py
AshishHirapara: done.

francesco-ooops:
@simahawk @legalsylvain is this ok for merge?

legalsylvain (approved):

legalsylvain:
/ocabot merge patch

Thanks.

PR #157, opened 2022-04-19
[13.0][IMP]base_user_role: Groups-roles navigation

GuillemCForgeFlow:
This PR improves the module base_user_role, implementing the navigation from the groups to the roles. It's not implemented the other way around as the groups can already be accessed through the main view of any role.

### DIFF
--- a/base_user_role/models/group.py
+++ b/base_user_role/models/group.py
@@ -1,0 +1,82 @@
+from odoo import api, fields, models
+
+
+class ResGroups(models.Model):
+    _inherit = "res.groups"
+
+    # The inverse field of the field group_id on the res.users.role model
+    # This field should be used a One2one relation as a role can only be
+    # represented by one group. It's declared as a One2many field as the
+    # inverse field on the res.users.role it's declared as a Many2one
+    role_id = fields.One2many(
+        comodel_name="res.users.role",
+        inverse_name="group_id",
+        help="Relation for the groups that represents a role",
+    )
+
+    role_ids = fields.Many2many(
+        comodel_name="res.users.role",
+        relation="res_groups_implied_roles_rel",
+        string="Roles",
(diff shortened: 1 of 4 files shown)


JordiMForgeFlow (approved):
Nice feature 🔝 ! Code + functional review LGTM 👍🏼

mariadforgeflow (approved):
LGTM! Code & functional review.

base_user_role/models/group.py
```diff
@@ -0,0 +1,77 @@
+from odoo import api, fields, models
+
+
+class ResGroups(models.Model):
+    _inherit = "res.groups"
+
+    role_id = fields.One2many(
+        comodel_name="res.users.role",
+        inverse_name="group_id",
+        help="Relation for the groups that are associated with a role",
```
sebalix:
```suggestion
        help="Relation for the groups that represents a role",
```
Because "associated with" could mean a group which is part of a role? A comment above saying that this field should be used as a "One2one" relation is welcome (that's why it is named like a Many2one I suppose).

GuillemCForgeFlow: Thanks for the suggestion! I've applied the changes according to what you were commenting. Hope that works now! :smile:

sebalix (approved):

sebalix:
@GuillemCForgeFlow oups, one last thing: can you remove the reference to the Odoo version in your commit message? (`[13.0]`)
When we backport or forward port commits from one branch to another this info becomes useless, and we avoid to change the commit messages when porting commits so we can compare them among different branches.

Side note: this doesn't apply for migration commits, as the changes they bring are really related to the current version of Odoo.

GuillemCForgeFlow:
> 

Perfect! I've changed the commit message! :+1:

PR #163, opened 2022-05-03
[ADD] pglogical

hbrunn:
this allows updates on a replicated (with pglogical) database without any database-foo for carrying schema changes to the replica(s)


hbrunn:
runboat build fixed in https://github.com/OCA/server-backend/pull/164

pankk (approved):

antonioburic (approved):

pglogical/hooks.py
```diff
@@ -0,0 +1,50 @@
...
+    if config["test_enable"]:
+        _logger.info("test mode enabled, not doing anything")
+        return
+
+    _logger.info("patching cursor to intercept ddl")
+    execute_org = Cursor.execute
+
+    def execute(self, query, params=None, log_exceptions=None):
+        """Wrap DDL in pglogical.replicate_ddl_command"""
+        if (query[:6] == "CREATE" or query[:5] == "ALTER" or query[:4] == "DROP") and (
```
antonioburic: Super nice work @hbrunn!

For recognizing DDL we could also use `sqlparse` to iterate through multiple statements if you think it's needed, [[get_type()](https://sqlparse.readthedocs.io/en/latest/analyzing/#sqlparse.sql.Statement.get_type)]. (though likely not necessary, I assume it should be rare to have multiple statements in)

hbrunn: I think we could use that in the second stage for detecting the commands we want to ignore. But the first conditional should be fast, this function is called *often*

StefanRijnhart: Maybe Odoo is consistent in its shouting style of SQL formatting, but the casing in migration scripts might vary. What do you think of applying `upper` to at least those first characters of the query that you are checking?

pglogical/hooks.py
```diff
@@ -0,0 +1,50 @@
...
+            # don't replicate constraints, triggers, indexes
+            "CONSTRAINT" not in query
+            and "TRIGGER" not in query
+            and "INDEX" not in query
```
StefanRijnhart: Here too, I am a little concerned about casing. However, when normalizing the rest of the query, there could be a false positive from table names like `text_index` or `update_trigger` so would you consider (if you have a match on the operation type above) to split the query by whitespace and check if you have a full word match on one of these terms rather than just a substring?

hbrunn: I think at this point we can use sqlparse as @antonioburic proposed above, then we're sure

pglogical/hooks.py
hbrunn: I added a somewhat different heuristics now, look at it more closely if it doesn't start with 'SELECT' - this way we don't impact select statements at all, and everything else gets the accurate treatment with the `sqlparse` lib.

StefanRijnhart: Would it be safe to exclude queries starting with UPDATE as well?

hbrunn: I think so, but do we expect a lot of updates that each take less than 100ms? Only then I think do we really have a relevant impact from the sql parsing

StefanRijnhart: That's a consideration of course. I'm not sure if it's going to be noticable, but it's a relatively large class of queries that we can save some overhead on.

hbrunn: I've added UPDATE

StefanRijnhart (approved):
Thanks, looking great!

hbrunn:
this shouldn't be merged just yet, as it failed an internal test

StefanRijnhart (approved):
Looking good, but let's block merging until we know this is working well in our setup.

antonioburic (approved):
great work again @hbrunn! 
(I initially thought we might need to be handle it/insert the schema from within the orm, which would be an issue for all the other statements, this will cover everything, nice!)

pankk (approved):

PR #164, opened 2022-05-03
[FIX] add external_dependencies_override for MySQLdb

hbrunn:
this makes the branch build on runboat, see https://pypi.org/project/setuptools-odoo/#controlling-setuptools-odoo-behaviour

### DIFF
--- a/setup/base_external_dbsource_mysql/setup.py
+++ b/setup/base_external_dbsource_mysql/setup.py
@@ -5,7 +5,7 @@ setuptools.setup(
     odoo_addon={
         'external_dependencies_override': {
             'python': {
-                'MySQLdb': 'mysqlclient',
+                'MySQLdb': 'mysqlclient==2.0.1',
             }
         },
     }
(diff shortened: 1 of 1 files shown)


pankk (approved):
It's green :smile:

setup/base_external_dbsource_mysql/setup.py
```diff
@@ -2,5 +2,11 @@
 
 setuptools.setup(
     setup_requires=['setuptools-odoo'],
-    odoo_addon=True,
+    odoo_addon={
+        'external_dependencies_override': {
+            'python': {
+                'MySQLdb': 'mysqlclient',
```
SirTakobi: The version is fixed in https://github.com/OCA/server-backend/blob/56fba17abc90466e9d9f31dfaa5578940cca7083/requirements.txt#L2 so I think the module might have some issues if we simply install the latest version.
I have proposed https://github.com/hbrunn/server-backend/pull/3 to pin the version can you have a look and maybe merge it?

SirTakobi (changes requested):
Thanks!
I was just going to make the same fix :smile:

setup/base_external_dbsource_mysql/setup.py
hbrunn: thanks

SirTakobi (approved):

StefanRijnhart (approved):

PR #165, opened 2022-05-05
[IMP] base_global_discount: Add discounted amount selection

SirTakobi:
Allow to apply global discounts on total

~~Runboat should be fixed after https://github.com/OCA/server-backend/pull/164~~

### DIFF
--- a/base_global_discount/__manifest__.py
+++ b/base_global_discount/__manifest__.py
@@ -1,5 +1,6 @@
 # Copyright 2019 Tecnativa S.L. - David Vidal
 # Copyright 2020 Xtendoo - Manuel Calero
+# Copyright 2022 Simone Rubino - TAKOBI
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 {
     'name': 'Base Global Discount',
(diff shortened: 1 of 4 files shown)


eLBati (approved):
Tested

SirTakobi:
@OCA/server-environment-maintainers can someone have a look?
Not very sure this is the correct team, please let me know if I have to mention something else, thanks!

SirTakobi:
@OCA/tools-maintainers can someone have a look? Thanks!

simahawk:
I'd ping an author instead (just look at the history). @chienandalu ping? :)

pedrobaeza:
Sorry, we are not using 12.0 anymore. Anyways, this is useless without the implementation in `account_global_discount`.

SirTakobi:
> Sorry, we are not using 12.0 anymore. Anyways, this is useless without the implementation in `account_global_discount`.

Thanks for the feedback!
This base module is already pretty much useless without the implementations in depending modules, it is mentioned in the README too:
> It doesn't do much for itself, so account_global_discount or purchase_global_discount should be installed to benefit from it.

(ref. https://github.com/OCA/server-backend/blob/3cd63d8a9b42ff4bcaca2c015737a7181b2da1e3/base_global_discount/README.rst#base-global-discount)
I just implemented this change here so I don't have to duplicate (N-plicate) the same code in N depending modules; right now this is only used in https://github.com/OCA/account-invoicing/pull/1150 but I suppose it might be used in `purchase_global_discount` (that I can't find right now) too.

pedrobaeza:
OK, but as stated, we don't work anymore with 12.0 at Tecnativa.

SirTakobi:
@OCA/tools-maintainers can someone have a look? Thanks!

chienandalu (approved):
Ok for me :+1:

PR #168, opened 2022-05-26
[15.0] FIX: `base_user_role` sudo

SilvioC2C:
1. FIX `base_user_role`: use `sudo()` in `user_ids` compute
When computing ``user_ids``, avoid AccessError due to multi-company rules if the compute is triggered by an admin user

2. FIX `base_user_role`: update tests

### DIFF
--- a/base_user_role/tests/test_user_role.py
+++ b/base_user_role/tests/test_user_role.py
@@ -3,34 +3,41 @@
 import datetime
 
 from odoo import fields
+from odoo.exceptions import AccessError
 from odoo.tests.common import TransactionCase
 
 
 class TestUserRole(TransactionCase):
-    def setUp(self):
-        super(TestUserRole, self).setUp()
-        self.user_model = self.env["res.users"]
-        self.role_model = self.env["res.users.role"]
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.env = cls.env(
+            context=dict(cls.env.context, tracking_disable=True, no_reset_password=True)
+        )
+        cls.user_model = cls.env["res.users"]
+        cls.role_model = cls.env["res.users.role"]
 
-        self.default_user = self.env.ref("base.default_user")
-        self.user_id = self.user_model.create(
+        cls.company1 = cls.env.ref("base.main_company")
+        cls.company2 = cls.env["res.company"].create({"name": "company2"})
+        cls.default_user = cls.env.ref("base.default_user")
+        cls.user_id = cls.user_model.create(
             {"name": "USER TEST (ROLES)", "login": "user_test_roles"}
         )
 
         # ROLE_1
-        self.group_user_id = self.env.ref("base.group_user")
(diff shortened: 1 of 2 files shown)


base_user_role/tests/test_user_role.py
```diff
@@ -3,51 +3,55 @@
 import datetime
 
 from odoo import fields
-from odoo.tests.common import TransactionCase
+from odoo.tests.common import SavepointCase
```
sebalix: `SavepointCase` has been deprecated in 15.0 and is replaced by `TransactionCase`: https://github.com/odoo/odoo/blob/15.0/odoo/tests/common.py#L746-L753=

Outside of this, ok to switch `setUp` to `setUpClass` but I can't see other fix, which issue you encountered exactly?

sebalix (changes requested):

base_user_role/tests/test_user_role.py
SilvioC2C: @sebalix when accessing the "Role" menu, the `base.res_users_rule` was triggered and an error was raised (because field `user_ids` might contain users with companies that are not chosen by the current user in the proper widget). I've added a proper test for it and force-pushed it a few seconds ago.

base_user_role/tests/test_user_role.py
```diff
@@ -169,3 +201,23 @@ def test_default_user_roles(self):
...
+        # Downgrade multicompany user 1 to common user
+        self.multicompany_user_1.write(
+            {"groups_id": [(6, 0, self.env.ref("base.group_user").ids)]}
+        )
+        # Check that the user cannot read multicompany data again since it lost
+        # its admin privileges
+        with self.assertRaisesRegex(
+            AccessError, "You are not allowed to access 'User role'"
+        ):
+            rule.read()
```
sebalix: Small typo to fix :)
```suggestion
        """Test AccessError when admin-like user accesses a role"""
        role = self.multicompany_role.with_user(self.multicompany_user_1)
        # Dummy read to check that multicompany user 1 has read access
        role.read()
        # Dummy read to check that multicompany user 1 has read access on the
        # whole role, even if it's using a different company than multicompany
        # user 2 (which is included in the role)
        role.with_context(allowed_company_ids=self.company1.ids).read()
        # Downgrade multicompany user 1 to common user
        self.multicompany_user_1.write(
            {"groups_id": [(6, 0, self.env.ref("base.group_user").ids)]}
        )
        # Check that the user cannot read multicompany data again since it lost
        # its admin privileges
        with self.assertRaisesRegex(
            AccessError, "You are not allowed to access 'User role'"
        ):
            role.read()
```

SilvioC2C: Sorry, it's Friday :smile: 
Fix'd!

sebalix (approved):

SilvioC2C:
@sebalix there's something wrong with the bot, it's still trying to merge...

PR #171, opened 2022-07-11
[15.0][FIX] UserWarning: Field res.groups.trans_parent_ids should be …

elmeriniemela:
…declared with recursive=True

Source for the warning: https://github.com/odoo/odoo/blob/15.0/odoo/fields.py#L720-L722


elmeriniemela:
@sebalix Can you merge this, please? It does not change the actual field definition, since Odoo sets recursive=True already before raising the warning (see https://github.com/odoo/odoo/blob/15.0/odoo/fields.py#L720-L722). All this PR does is remove the annoying warning from logs.

sebalix (approved):
LG, this warning was annoying indeed

PR #178, opened 2022-08-18
[FIX][12.0] pglogical: don't replicate aggregate functions

hbrunn:
this fixes custom aggregates by not replicating them


StefanRijnhart (approved):
Thanks! Fun fact for other reviewers: this was uncovered by https://github.com/OCA/account-financial-reporting/blob/12.0/account_financial_report/report/trial_balance.py#L486-L490.

pankk (approved):

antonioburic (approved):

PR #190, opened 2022-10-18
[DOC] base_user_role: set development status to 'Production/Stable'

sebalix:
This module is used since a long time over quite a lot of Odoo versions, time to set a convenient development status.


sebalix:
@sbidoul @lmignon as you are using this module, OK for this switch?

sbidoul (commented):
Yes, this one is definitely production-ready :)

sbidoul (approved):

PR #192, opened 2022-10-18
[15.0][ADD] base_external_dbsource_sap_hana: New module to connect to SAP Hana using Alchemy

sergio-teruel:
cc @Tecnativa TT38753

ping @carlosdauden @chienandalu


chienandalu (approved):
Looks good to me, although as in other modules like this one, Wouldn't an sql_alchemy connector be enough for this kind of connectors?

PR #202, opened 2022-12-29
[16.0][MIG] base_global_discount: 16.0

ferran-S73:


base_global_discount/models/res_partner.py
```diff
@@ -0,0 +1,38 @@
...
+        relation="supplier_global_discount_rel",
+        column1="partner_id",
+        column2="global_discount_id",
+        string="Purchase Global Discounts",
+        domain=[("discount_scope", "=", "purchase")],
+    )
+    # HACK: Looks like UI doesn't behave well with Many2many fields and
+    # negative groups when the same field is shown. In this case, we want to
+    # show the readonly version to any not in the global discount group.
+    # TODO: Check in future versions if it's fixed
```
rousseldenis: @ferran-S73 Can you check this ?

ferran-S73: @rousseldenis I did a couple of tests and it appears to be working fine with just one field

rousseldenis (approved):
Code review

PR #212, opened 2023-04-14
[REF] base_external_dbsource[_sqlite]: modernize, make it sqlalchemy-wide

yajo:
[REF] base_external_dbsource: modernize

I dropped support for deprecated APIs from v10:

- Remove `conn_open`.
- Drop support for `sqlquery` and `sqlparams` kwargs in `execute`.

I made all methods private:

- Some methods were interacting with rich data types. It made no sense to make them publicly reachable by RPC, where only standard data types are allowed.
- Some other methods allowed to execute uncontrolled queries on any data source by any RPC call. These calls were still secured by access rules, but since there's no UI for making them, it seemed like the only purpose for these methods was to be abused. Downstream modules that provide a UI on top of this module should implement controlled public endpoints instead.

I removed some nonsense:

- The standard password string `PWD=%s;` was arbitrarily added to any connector that didn't specify one. This makes no sense. That parameter might produce an error on another connector's query string. Besides, there are query strings which have no password, and that's on purpose because it's not required (example: sqlite). Now, that's only added if the current connector is PostgreSQL (where it makes sense).
- The `CONNECTORS` class pseudo-constant was a source of monkey-patching practices that are now removed in migrations of `base_external_dbsource_[sqlite,mssql]`. To prevent from further monkey-patching, I removed it.

I adopted the module.

---

[REF] base_external_dbsource_sqlite: make it sqlalchemy-wide instead

It turns out that there's a lot of modules that extend this one because they're based on sqlalchemy... that are not needed at all.

Once you select the connection engine to use, you have to provide a full sqlalchemy query string. So, you can actually select sqlite as the engine, but provide `mssql+pyodbc://` if you want (for example). It just works, because sqlalchemy is itself an abstraction layer on top of db connectors.

This module should be named instead `base_external_dbsource_sqlalchemy`, but now that it's got the sqlite name, I think it's just easier to rename the parts that are exposed to the users:

- Rename the connector to SQLAlchemy.
- Rename the methods accordingly (also according to refactor on `base_external_dbsource`).
- Update instructions.

This will probably make it more clear now that modules such as `base_external_dbsource_[mssql,odbc]` are really unnecessary.

I adopted this module too.

---

[DEL] base_external_dbsource_mssql: non-installable

This module is no longer needed, according to its dependencies refactors. Remove it.

---
@moduon MT-2666


base_external_dbsource/models/base_external_dbsource.py
```diff
@@ -34,11 +34,10 @@ class BaseExternalDbsource(models.Model):
...
 
-    CONNECTORS = [("postgresql", "PostgreSQL")]
     # This is appended to the conn string if pass declared but not detected.
     # Children should declare PWD_STRING_CONNECTOR (such as PWD_STRING_FBD)
     #   to allow for override.
-    PWD_STRING = "PWD=%s;"
+    PWD_STRING_POSTGRESQL = "PWD=%s;"
 
     name = fields.Char("Datasource name", required=True)
     conn_string = fields.Text(
```
Shide, line 38: Add a hint in the help to indicate that `%s` is where the password is going to be.

base_external_dbsource/models/base_external_dbsource.py
```diff
@@ -34,11 +34,10 @@ class BaseExternalDbsource(models.Model):
     _name = "base.external.dbsource"
     _description = "External Database Sources"
 
-    CONNECTORS = [("postgresql", "PostgreSQL")]
     # This is appended to the conn string if pass declared but not detected.
     # Children should declare PWD_STRING_CONNECTOR (such as PWD_STRING_FBD)
     #   to allow for override.
-    PWD_STRING = "PWD=%s;"
+    PWD_STRING_POSTGRESQL = "PWD=%s;"
```
Shide: I think this can be deleted. 
The `conn_string` field should be the responsible of provide a correct connection string to format the password.

base_external_dbsource/models/base_external_dbsource.py
```diff
@@ -68,36 +67,39 @@ class BaseExternalDbsource(models.Model):
...
+                            record,
+                            "PWD_STRING_%s" % record.connector.upper(),
+                        )
+                    except AttributeError:
+                        _logger.info("Not altering query string for %s.", record)
+                    else:
+                        record.conn_string += pwd_string
                 record.conn_string_full = record.conn_string % record.password
             else:
                 record.conn_string_full = record.conn_string
```
Shide: If `PWD_STRING_POSTGRESQL` variable is deleted, here we just go for 
```python
conn_string_full = record.conn_string
try:
    conn_string_full = record.conn_string % (record.password or '')
except TypeError:
    _logger.info("Password not needed")
record.conn_string_full = conn_string_full
```

base_external_dbsource/models/base_external_dbsource.py
```diff
@@ -296,19 +285,6 @@ def _execute_generic(self, query, params, metadata):
...
-        the adapter type.
-
-        Deprecate:
-            This method has been replaced with ``connection_open``.
-        """
-
-        with self.connection_open() as connection:
-            return connection
-
     def _get_adapter_method(self, method_prefix):
```
Shide, line 273: Why not remove this?
Only connection_open and connection_close are needed on every connection. Just use with normal inheritance and forget adapters...

Every module must override the connection_open and connection close as this:

```python
def connection_open(self):
    self.ensure_one()
    if self.connector == "postgresql":
        return my_postgresql_library.connect(self.conn_string_full).open()  # Return an opened connection
    return super().connection_open()
```

or just use contextmanagers like Odoo does (it's not the exact code, but a hint):

```python
@contextmanager
def connect(self):
    self.ensure_one()
    if self.connector != "postgresql":
        return super().connect()  # Maybe it's a yield... ?
    connection = my_postgresql_library.connect(self.conn_string_full)
    try:
        yield connection.open()  # Return an opened connection
    finally:
        connection.close()
```

and use it like this:
```python
# We are on connection model
def do_something(self):
    self.ensure_one()
    with self.connect() as connection:  # Auto opened
        usable_records = connection.browse("something")
    # Auto closed when exit `with`
```

base_external_dbsource/models/base_external_dbsource.py
yajo: It can't really be that simple because, if there's no `%s` in `conn_string` and I format it, Python will raise an exception. However, I'll simplify it.

base_external_dbsource/models/base_external_dbsource.py
yajo: This suggestion makes total sense. 

However, the current system doesn't make end uses of the base module more complex (because all the `_get_adapter_method()` dance is done directly in the base module, and only there), and it doesn't make sub-implementations particularly complex (see [this example](https://github.com/OCA/server-backend/blob/427c6cccc84d605cec2f8a5b115f68807cecc292/base_external_dbsource_sqlite/models/base_external_dbsource.py); it's quite simple anyway).

In this refactor, I'm changing stuff that actually cause problems. But following this suggestion would make it a deeper change for something that, although not very ergonomic, isn't causing problems. Thus, I think it's better to keep it as it is, to minimize the refactor here.

yajo:
All attended.

Shide (approved):

dalonsod:
@yajo could you reopen this? I would migrate `base_external_dbsource[_sqlite]` to v17, and this refactoring should be available first IMO. Thanks!

cc @Anxo82

yajo:
I won't reopen it, because we don't use these modules anymore, but please feel free to cherry-pick the commits from this PR and build yours on top of them.

dalonsod:
Thanks @yajo !

dalonsod:
@yajo ouch!!! You've deleted your origin branch, so we can't access it for cherry-pick those commits

imagen

Could you restore it or is it been permanently deleted?

yajo:
Just download the ref:

    git pull origin refs/pull/212/head

Or apply the patch:

    curl -sSL https://github.com/OCA/server-backend/pull/212.patch | git am -3

PR #214, opened 2023-04-26
[ADD] base_portal_type

hbrunn:
This module allows administrators to configure groups for portal users

- create a group of category ``base_portal_type.category_portal_type``
- observe the group shows up on a user form for users of type `Portal`
- now you can use this group in your frontend templates and code to make some features accessible to this group


base_portal_type/models/res_groups.py
```diff
@@ -0,0 +1,92 @@
...
+                                    "invisible": [
+                                        (user_type_field, "!=", portal_group.id),
+                                    ],
+                                }
+                            ),
+                        },
+                    )
+                )
+        view_context = dict(self.env.context, lang=None)
+        view_context.pop("install_filename", None)
```
antonioburic, line 73: I'm not sure I understand this fully, we do this in order not to set the filename as it's always recomputed?

antonioburic (approved):
looks good to me, interesting idea!

base_portal_type/models/res_groups.py
hbrunn: it avoids a crash in https://github.com/OCA/OCB/blob/16.0/odoo/addons/base/models/ir_ui_view.py#L336 as this is set to a non-existent path at that point

StefanRijnhart (approved):
Code is great, as always :smiley:, and it seems to work well. I'm happy to see that Odoo prevents you from changing the user type by accidentally inheriting from the internal users group in any portal group.
I think the module description is a bit sparse. Would you consider describing the problem (something along the lines of "Portal users in standard Odoo cannot be divided into user groups"). before you describe the solution?

StefanRijnhart:
@hbrunn please have a look:
```
2023-06-05 20:50:33,589 242 ERROR odoo odoo.addons.base_portal_type.tests.test_base_portal_type: FAIL: TestBasePortalGroup.test_portal_group_view
  File "/__w/server-backend/server-backend/base_portal_type/tests/test_base_portal_type.py", line 18, in test_portal_group_view
    self.assertEqual(
AssertionError: 3 != 2
```

hbrunn:
sorry everyone for following up so late, this one slipped me somehow. Added a fixup commit for your convenience, will squash when the first reviewer agrees.

changing this to 3 is not laziness but the result of studying https://github.com/oca/ocb/commit/9deb1e6aa902a09c4fd6a91ecbd34a89b9c98f98 which is why the expected amount of this type of group is 3 by now.

StefanRijnhart:
Thanks, please go ahead with the squash!

hbrunn:
@StefanRijnhart thank you, done

PR #215, opened 2023-06-06
[FIX] AttributeError: 'str' object has no attribute '_execute_on_connection'

Guenzn:
Deprecated since version 2.0: passing a string to [Connection.execute()](https://docs.sqlalchemy.org/en/14/core/connections.html#sqlalchemy.engine.Connection.execute) is deprecated and will be removed in version 2.0. Use the [text()](https://docs.sqlalchemy.org/en/14/core/sqlelement.html#sqlalchemy.sql.expression.text) construct with [Connection.execute()](https://docs.sqlalchemy.org/en/14/core/connections.html#sqlalchemy.engine.Connection.execute), or the [Connection.exec_driver_sql()](https://docs.sqlalchemy.org/en/14/core/connections.html#sqlalchemy.engine.Connection.exec_driver_sql) method to invoke a driver-level SQL string.

https://docs.sqlalchemy.org/en/14/core/connections.html#sqlalchemy.engine.Connection.execute

### DIFF
--- a/base_external_dbsource_sqlite/models/base_external_dbsource.py
+++ b/base_external_dbsource_sqlite/models/base_external_dbsource.py
@@ -30,12 +30,13 @@ class BaseExternalDbsource(models.Model):
 
     def _execute_sqlalchemy(self, sqlquery, sqlparams, metadata):
         rows, cols = list(), list()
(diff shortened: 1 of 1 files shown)


wildi1:
+1 Please merge

Issue #218, opened 2023-06-17
base_user_role: create role from existing user


gpothier:
I created a module that extends `base_user_role` to add a wizard to create a new role based on an existing user. 

![image](https://github.com/OCA/server-backend/assets/629806/1f70a4fc-da7d-45b0-8fb6-b327e6b9058b)

It basically adds a wizard that permits to create a role that contains the same groups as the selected user.

![image](https://github.com/OCA/server-backend/assets/629806/c83976c8-5337-466c-959c-70dc589dc775)

I'd like to contribute it to OCA. I think it is pretty useful for people who start using `base_user_role` after having already created several users. So:

- Would it be of interest to the community?
- Would it be best to keep it as a separate module, or rather to include the wizard in `base_user_role` module? My personal preference would be to include it in the existing module, as it is very small.

I have a version for Odoo v8 and another for Odoo v16. There are no functional differences between them.

legalsylvain:
Hi. Very Nice feature indeed !
As it is only a transient modèle, i think this feature should go in the main module. 

The contribution should be done on récent version. 8 is too old imo. 

Regards.

gpothier:
Closing as PR #219 was merged

PR #219, opened 2023-06-20
Add "Create role from user" feature

gpothier:
Proposal to close issue #218. This adds a wizard to create a new role based on the groups of a specific user.

### DIFF
--- a/base_user_role/README.rst
+++ b/base_user_role/README.rst
@@ -78,6 +78,10 @@ Add users (with dates or not):
    :width: 80 %
    :align: center
 
+Instead of creating roles from scratch, it is possible to create a role
+based on the groups of an existing user: select or open the user and choose
+"Create role from user" in the action menu.
+
 Usage
 =====
 
@@ -116,6 +120,7 @@ Contributors
 * Kevin Khao <kevin.khao@akretion.com>
 * Tatiana Deribina <tatiana.deribina@sprintit.fi> (https://sprintit.fi)
 * Guillem Casassas <guillem.casassas@forgeflow.com>
+* Guillaume Pothier <gpothier@caligrafix.cl>
 
 Do not contact contributors directly about support or help with technical issues.
 
--- a/base_user_role/README.rst
+++ b/base_user_role/README.rst
@@ -23,7 +23,7 @@ User roles
     :target: https://runbot.odoo-community.org/runbot/253/16.0
     :alt: Try me on Runbot
 
-|badge1| |badge2| |badge3| |badge4| |badge5| 
+|badge1| |badge2| |badge3| |badge4| |badge5|
 
 This module was written to extend the standard functionality regarding users
 and groups management.
@@ -78,10 +78,6 @@ Add users (with dates or not):
    :width: 80 %
    :align: center
 
-Instead of creating roles from scratch, it is possible to create a role
-based on the groups of an existing user: select or open the user and choose
-"Create role from user" in the action menu.
-
 Usage
 =====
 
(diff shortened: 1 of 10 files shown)


base_user_role/wizard/create_from_user.py
```diff
@@ -0,0 +1,46 @@
+from odoo import models, fields
+
+
+class CreateFromUserWizard(models.TransientModel):
+    _name = "base_user_role.create_from_user_wizard"
+    _description = "Create role from user wizard"
+
+    name = fields.Char("Name")
```
legalsylvain:
```suggestion
    name = fields.Char(required=True)
```

- since recent version, string is not guessed from the field, so it's unnecessary to set it, if it is not different.
- name is required, otherwise, the call of the function ``create_from_user`` will fail is name is not set.

base_user_role/wizard/create_from_user.py
```diff
@@ -0,0 +1,46 @@
+from odoo import models, fields
+
+
+class CreateFromUserWizard(models.TransientModel):
+    _name = "base_user_role.create_from_user_wizard"
```
legalsylvain:
```suggestion
class WizardCreateRoleFromUser(models.TransientModel):
    _name = "wizard.create.role.from.user"
```

proposing name more simple.

base_user_role/README.rst
```diff
@@ -116,6 +120,7 @@ Contributors
 * Kevin Khao <kevin.khao@akretion.com>
 * Tatiana Deribina <tatiana.deribina@sprintit.fi> (https://sprintit.fi)
 * Guillem Casassas <guillem.casassas@forgeflow.com>
+* Guillaume Pothier <gpothier@caligrafix.cl>
```
legalsylvain: You should not update this file (README.rst) as it is an autogenerated file, updated when the PR is merged.

you can change fragments present in the readme folder.

base_user_role/wizard/create_from_user.py
```diff
@@ -0,0 +1,46 @@
...
+    _description = "Create role from user wizard"
+
+    name = fields.Char("Name")
+    assign_to_user = fields.Boolean("Assign to user", default=True)
+
```
legalsylvain, line 10:
```suggestion
    user_id = fields.Many2one(
        comodel_name="res.users",
        required=True,
        readonly=True,
        default=lambda x: x._default_user_id()
    )
    
    def _default_user_id(self):
        user_ids = self.env.context.get("active_ids", [])
        if user_ids:
             return user_ids[0]
        return False

```
This change allow to display the field in the view of the wizard.

(also it allow to have a more simple create_from_user function as the user will be available with self._user_id.

legalsylvain (changes requested):
Hi @gpothier. Thanks a lot for your first contribution in the OCA.

I put some suggestions inline. feel free to accept it if you think it's relevant.

Note : pre-commit check are failing.

you can take a look on the red CI in the "Detail" link
image

You can also run locally ``pre-commit run -a`` to try to autofix the error.

Nice to have : could you add a test for this new feature ? 

regards.

base_user_role/wizard/create_from_user.py
gpothier: I'd rather skip this one if you are ok with this. As far as I can tell it is not common for wizards that apply on selected records to display the records they are being applied on. On the other hand, the simplification in `create_from_user` is not significant IMHO, and comes at the cost of these several suggested additional lines of code.

gpothier:
Hi @legalsylvain, thanks a lot for the reviews. I think most of your concerns have now been addressed.

gpothier:
Ah, I just realized maybe it is necessary to bump the version number? Or is it done automatically?

legalsylvain:
> Ah, I just realized maybe it is necessary to bump the version number? Or is it done automatically?

no need to bump version. It is done automatically when merging, by the oca bot.

legalsylvain (approved):
thanks !

hbrunn (approved):

legalsylvain (approved):
/ocabot merge minor

PR #234, opened 2023-08-21
[FIX][IMP]requirements for bookworm

flotho:


DorianMAG (approved):
Tested in debian with python 3.11
LGTM
Thx @flotho

JulienMartinez (approved):
Tested in local
LGTM

PR #235, opened 2023-08-30
[16.0][MIG] base_external_dbsource_mysql

dalonsod:
Standard migration


dalonsod:
In this case tests don't cover real life examples... And locally tested I've found some problems:

- I'm using Python 3.8.
- With current SQLAlchemy python package (2.0.20), making an `execute()` command an error is raised:
```
  File "/home/xxx/.venv38/lib/python3.8/site-packages/sqlalchemy/engine/base.py", line 1408, in execute
    meth = statement._execute_on_connection
AttributeError: 'str' object has no attribute '_execute_on_connection'
```

  This error seems to be due to SQLAlchemy version (see https://stackoverflow.com/a/75464429), that suggests downgrading library or using a wrap function for this new version,
- With downgrade option I've successfully performed an `execute()`, but a warning is obtained: 
`2023-08-30 14:58:33,303 18329 WARNING odoo_16_oca01 py.warnings: /home/xxx/server-backend/base_external_dbsource_mysql/models/base_external_dbsource.py:31: RemovedIn20Warning: Deprecated API features detected! These feature(s) are not compatible with SQLAlchemy 2.0. To prevent incompatible upgrades prior to updating applications, ensure requirements files are pinned to "sqlalchemy<2.0". Set environment variable SQLALCHEMY_WARN_20=1 to show all deprecation warnings.  Set environment variable SQLALCHEMY_SILENCE_UBER_WARNING=1 to silence this message. (Background on SQLAlchemy 2.0 at: https://sqlalche.me/e/b8d9)`

I'm little bit confused, but in this situation, I've only see two possible scenarios:
- (1) Explicity use `sqlalchemy<2.0` in `requirements.txt`, or
- (2) Adapt base addon `base_external_dbsource` to new `sqlalchemy` >= 2.0 version, chaning line

  https://github.com/OCA/server-backend/blob/8292cc6aab19cc1ce5b4fd10836438ad0b230323/base_external_dbsource/models/base_external_dbsource.py#L161

 to `rows, cols = method(sqlalchemy.text(query), execute_params, metadata)`

Both SQLite and MSSQL addons are already merged... @anddago78 are you currently using MSSQL addon, and is it working with SQLAlchemy library newer versions?

cc @yajo

yajo:
Hello! We started using those addons but finally stopped on that because the way they're designed makes no sense.

1. The base addon only connects to postgres, without SQLAlchemy. Sorta kinda OK, but in reality, the base module could perfectly include SQLAlchemy, which can be used also to connect to postgres. Thus we'd only need one implementation for postgres, instead of 2.
2. The sqlite addon in reality could be named `external_dbsource_sqlalchemy` (or just `base_external_dbsource`; see point 1). By installing SQLAlchemy along with Odoo, you have out of the box support for sqlite and postgres. Thus, no sense to split in 2 modules.
3. Any other modules are also unnecessary. SQLAlchemy does everything by itself. You just need to install the extra python package required to connect to the new backend, and the sqlite module is enough for you.
4. Some nonsense accumulated over the years on the code.

Fixing all of this was too much burden for our use case, so we just stopped using these modules.

So, chances are that if you install the sqlite addon + the mysql connector library, you don't really need this module.

dalonsod:
Hello @yajo thanks for everything, just tested my custom MySQL connection with only `base_external_dbsource_sqlite` module and worked fine!

The SQLAlchemy python package version issue should be ported to another thread, but if you have a point of view of this and needs a new PR, I'd do it myself. I've noticed that such problem is still present: with modern python version installed SQLAlchemy library version seems to be not compatible with `base_external_dbsource` code. Should be freezed `sqlachemy` version in base module, or updated the code in order to adapt it to newer versions?

yajo:
Well, the dependencies are gonna be a problem:
- [Odoo requires greenlet==0.4.17 and gevent==20.9.0](https://github.com/odoo/odoo/blob/c9e5e6437453978bded2b1d4eb79b69df83cda77/requirements.txt#L16)
- [SQLAlchemy requires greenlet!=0.4.17](https://github.com/sqlalchemy/sqlalchemy/blob/228490ead7048f2e558c25b0f055bdb952272ec4/setup.cfg#L40)
- [Gevent requires strict ABI compatibility with greenlet](https://github.com/gevent/gevent/blob/1e412d35526183b26c1abf2eb658cbef661f5f70/setup.py#L207-L219)

So, if you pin SQLAlchemy, you'll have to pin Greenlet and Gevent too, to versions explicitly different to those pinned by Odoo. You have surgery work ahead! 😉

base_external_dbsource_mysql/models/base_external_dbsource.py
```diff
@@ -0,0 +1,37 @@
...
+        rows, cols = list(), list()
+        for record in self:
+            with record.connection_open() as connection:
+                if sqlparams is None:
+                    cur = connection.execute(sqlquery)
+                else:
+                    cur = connection.execute(sqlquery, sqlparams)
+                if metadata:
+                    cols = list(cur.keys())
+                rows = [r for r in cur]
```
mrm-david, line 36: This is going to raise an exception if a query that doesn't return rows is executed.
`rows = [r for r in cur] if cur.returns_rows else []` fixes this

dalonsod: Thank you @mrm-david this could be proposed as a PR for `base_external_dbsource_sqlite` addon (which has the same issue and, as @yajo said, could be used instead of this MySQL addon). Could you make the PR?

mrm-david: @dalonsod Would I need to write a unit test ? I have sadly very little experience with those.

dalonsod: regarding to test coding, these addons are very particular, as mock objects are used to emulate tests, see 

https://github.com/solvosci/server-backend/blob/33de09e95bb8d8ef3340d50722b8777b69cfe630/base_external_dbsource_sqlite/tests/test_base_external_dbsource.py#L32-L48

Thus, your change won't imply tests update if I'm right.

mrm-david: I've submitted https://github.com/OCA/server-backend/pull/243 in case you want to add a review.

dreispt (approved):

PR #236, opened 2023-09-11
[14.0] base_user_role - add (copy) in order to allow the duplicate

ghost:
Without this patch, while duplicating, we have a constraint error on the "res.groups" model (name_uniq)

### DIFF
--- a/base_user_role/models/role.py
+++ b/base_user_role/models/role.py
@@ -3,7 +3,7 @@
 import datetime
 import logging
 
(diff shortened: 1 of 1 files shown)


legalsylvain (approved):
Thanks for this patch.

/ocabot merge patch

Could you port it to the more recent version ? (16.0)

thanks.

ghost:
@legalsylvain You will have the PRs for 15.0 and 16.0 in a few minutes

PR #139, opened 2021-11-11
[MIG] base_user_role_company: Migration to 15.0

Tatider:
Depends on [base_user_role](https://github.com/OCA/server-backend/pull/138) migration.

PR #241, opened 2023-09-27
[15.0][MIG] base_user_role_company

hbrunn:
this is a backport from v16 as it contains patches the previous v15 PRs (https://github.com/OCA/server-backend/pull/139, https://github.com/OCA/server-backend/pull/187) didn't


base_user_role_company/README.rst
```diff
@@ -0,0 +1,119 @@
...
+.. |badge5| image:: https://img.shields.io/badge/runboat-Try%20me-875A7B.png
+    :target: https://runboat.odoo-community.org/builds?repo=OCA/server-backend&target_branch=16.0
+    :alt: Try me on Runboat
+
+|badge1| |badge2| |badge3| |badge4| |badge5|
+
+Enable User Roles depending on the Companies selected.
+
+A company specific Role will only be enabled
+if it is set for **all** the currently selected companies.
```
hbrunn, line 34: @dreispt this isn't true since at least v14. To me it would make sense to adapt the code accordingly, or do you as maintainer say better adapt the readme?

JordiBForgeFlow (approved):

JordiBForgeFlow:
@hbrunn Looks good. Perhaps a small glitch (not associated with the migration) is that when you access to the users associated to a role, from the role, the company assigned to the user is not visible.

image

It is only visible from the user:

image

hbrunn:
@JordiBForgeFlow I think I can slip this into this PR if nobody objects?

Do you have an opinion about https://github.com/OCA/server-backend/pull/241#pullrequestreview-1646067495 ?

JordiBForgeFlow:
@hbrunn I would add the change as a separate commit, so that we can fwp and backport.

base_user_role_company/README.rst
JordiBForgeFlow: @hbrunn  agree that the current README does not reflect what the code does.

The current code will poses a threat from a security standpoint. If you are assigned to Sales Manager role in company A only, you can still access to all sales manager functions in company B by just enabling this company together with company A.

I think that it's necessary to adapt the code to the README.

JordiBForgeFlow: @hbrunn  agree that the current README does not reflect what the code does.

The current code will poses a threat from a security standpoint. If you are assigned to Sales Manager role in company A only, you can still access to all sales manager functions in company B by just enabling this company together with company A.

I think that it's necessary to adapt the code to the README.

sebalix: Is there some news about this?

hbrunn:
added the company column on the user tab in roles in https://github.com/OCA/server-backend/pull/241/commits/a2e03bf0f7f33c2741d87335d93009c155bbab4b, and fixed the behavior for multiple selected companies in https://github.com/OCA/server-backend/pull/241/commits/28c9cec393e1b8f430c28d096a0a393abffc1079

base_user_role_company/README.rst
hbrunn: fixed this right now

hbrunn: @JordiBForgeFlow @sebalix @dreispt but wait, if this is supposed to be a security feature (as opposed to convenience as I saw it up to now), then the module needs to be revamped entirely:

We set the groups only on authentication or when get_session_info is called, that means if I authenticate, select company1 only, I'll have groups specific to company1.

For people just clicking things in the webclient, that's fine. But if I now craft a request for this session with a context setting allowed companies to company1,company2, this request runs with the groups specific to company1, but possibly on data of company2.

As we don't want to update groups for *every* request, and we can't rewrite `base_user_role` to not work by rewriting groups, I think we'll have to save somewhere server side for which companies the last update of groups has taken place, and restrict to/raise if we get allowed companies not in this list? With the current approach you can't be sensibly in company1 with one tab and company2 with another tab and the same user anyways, so this wouldn't really make things worse?

JordiBForgeFlow: With the current data model, I don't see how to fix this issue. Seems as if we'd want to store, a model res.user.group.company that stores for which companies is a user assigned to a group..

hbrunn: I think introducing a field `last_cids` on res.users that we write every time we rewrite groups should work. Then we can restrict the incoming allowed companies to the intersection with `last_cids`, and the company switcher then has to call a controller that does updating last_cids + group update in one transaction.

With this we should be able to guarantee you never have groups assigned for companies that shouldn't.

dgiraldoc2c: Hello here!
I work for an Odoo integrator. I think this module could be quite useful for some clients. I subscribe to the analysis that it doesn't do what it's expected to as soon it is considered as a security module (which is how I see it and how I think it could be useful).
I'm a functional guy so I'm probably missing many technical aspects. Anyways, here is my take about what the module should do : 
- Regarding the Menus / Views / Access Rights, consider the intersection of what the companies ticked in the selector give access to (through the groups associated to the roles and their inherited groups). Example with Access Rights, if you have access to those two groups through companies ticked in the selector : 
image
image
=> Then you should only get read and write access to sales orders and sale orders lines
- Regarding the record rules : apply an "AND" rule. In the example below we have 2 groups of record rules : 
image
image
=> In this example, since we have a record rule restricting the modification to one's own sales and the other to the sales in "draft" state (quotations), if the companies in the selector give access to those two groups, the modification should only be allowed for one's own sales in "draft" state.

What's your opinion ? Happy to discuss this topic with you !

hbrunn: what you describe is changing how roles work in the first place. That would involve changing the base module, so very much out of scope here.

dgiraldoc2c: Hello @hbrunn ,
Thanks for reading my post and coming back to me!
Do you have other solutions in mind ?

hbrunn: what I wrote [above](https://github.com/OCA/server-backend/pull/241#discussion_r1495217717), but that also won't happen in this PR I think

PR #147, opened 2022-02-15
[14.0][FIX] base_user_role_company: roles not properly applied on login

dreispt:

PR #242, opened 2023-10-05
[14.0][IMP/FIX] base_user_role_company - re-add logic around multiple active companies

ghost:
Since this PR https://github.com/OCA/server-backend/pull/147, the logic when you have several companies active at the same time is broken because only the main/active company is considered (And then inconsistent with the README of the module)

Imagine the following scenario where all companies below are active :

- COMPANY A (Active/Main company): Accounting Manager
- COMPANY B: Purchase Manager
- COMPANY C: Inventory Manager
- ALL COMPANY : Seller

Without the patch:  The user will be Seller and Account Manager on all companies
With the patch: The user will be only Seller on all companies

Another situation is also managed. Having the same role for several companies.

### DIFF
--- a/base_user_role_company/models/ir_http.py
+++ b/base_user_role_company/models/ir_http.py
@@ -18,6 +18,5 @@ class IrHttp(models.AbstractModel):
         if self.env.user.role_line_ids:
             cids_str = request.httprequest.cookies.get("cids", str(self.env.company.id))
             cids = [int(cid) for cid in cids_str.split(",")]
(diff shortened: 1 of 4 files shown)


ghost:
@OCA/tools-maintainers @dreispt Can you have a look at this PR ?

dreispt (approved):

dreispt:
I did not test it, but I agree with the implementation goals and code review looks good.

PR #243, opened 2023-10-09
Prevent exception if the query doesn't return rows

mrm-david:
Trying to get rows from the cursor when the query doesn't return rows will raise an exception.

```
sqlalchemy.exc.ResourceClosedError: This result object does not return rows. It has been closed automatically.
```
An example such request would for, by example, in mysql `SET some_variable = some_value`


legalsylvain (approved):

mrm-david:
Note: I have no idea how I could write a unit test for this

mrm-david:
Also should I fix the trailing space in the comment to have 1 more check that pass, or leave it as is (automatically corrected)

legalsylvain:
Run
 pre-commit run -a 
Should fix the syntax error.

dreispt (approved):

dreispt:
Superseded by #274

PR #250, opened 2023-11-15
[15.0][IMP] base_user_role: Add role_ids column to tree view.

amh-mw:
The role_ids field is already searchable in list view, now it is also displayed.

Screenshot 2023-11-15 at 10 49 28 AM

### DIFF
--- a/base_user_role/views/user.xml
+++ b/base_user_role/views/user.xml
@@ -39,4 +39,13 @@
             </field>
         </field>
     </record>
+    <record id="view_res_users_tree_inherit" model="ir.ui.view">
+        <field name="model">res.users</field>
+        <field name="inherit_id" ref="base.view_users_tree" />
+        <field name="arch" type="xml">
+            <field name="company_id" position="before">
+                <field name="role_ids" widget="many2many_tags" />
+            </field>
+        </field>
+    </record>
 </odoo>


base_user_role/__manifest__.py
```diff
@@ -4,7 +4,7 @@
 
 {
     "name": "User roles",
-    "version": "15.0.0.3.4",
+    "version": "15.0.0.3.5",
```
legalsylvain: no need to bump the version, it will be done automatically when merging. 
(and it can avoid conflict).

base_user_role/views/user.xml
```diff
@@ -39,4 +39,14 @@
             </field>
         </field>
     </record>
+    <record id="view_res_users_tree_inherit" model="ir.ui.view">
+        <field name="name">res.users.tree.inherit</field>
```
legalsylvain: not blocking point : FYI this field is useless. (and is autogenerated, if not set).

base_user_role/views/user.xml
```diff
@@ -39,4 +39,14 @@
...
+        <field name="model">res.users</field>
+        <field name="inherit_id" ref="base.view_users_tree" />
+        <field name="arch" type="xml">
+            <field name="company_id" position="after">
```
legalsylvain:
```suggestion
            <field name="company_id" position="before">
```

In odoo, the company_id field is usually the last column of the tree view. So better to keep that design, and set the value before.

legalsylvain (commented):
some minors remarks. 
Otherwise, LGTM. Thanks !

base_user_role/__manifest__.py
amh-mw: That's a good trick. I'll have to see about adding it to the pre-commit of my own modules.

legalsylvain:
Thanks ! 

/ocabot merge patch

PR #251, opened 2023-11-21
[15.0][IMP] base_user_role: users validation

polqubiq:
Add a restriction to make sure the Administrator user isn't included in the Users of the Roles.
This helps avoid problems that come up when you add the admin user there, like replacing all permissions based on the role's groups.

### DIFF
--- a/base_user_role/models/role.py
+++ b/base_user_role/models/role.py
@@ -4,6 +4,7 @@ import datetime
 import logging
 
 from odoo import SUPERUSER_ID, _, api, fields, models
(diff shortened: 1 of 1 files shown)


legalsylvain (changes requested):
set a role to the admin user can be a valid use case.

Such extra restriction should be done in a custom module IMO.

polqubiq:
> set a role to the admin user can be a valid use case.
> 
> Such extra restriction should be done in a custom module IMO.

Okay, it makes sense.

PR #258, opened 2023-12-23
[16.0] mig server_action_navigate (from 14.0)

legalsylvain:
### Description

This module provides a generic tool to have the possibility to see a list of records associated to a given selection of records. This is basically the UI version of ``recordset.mapped('field1.field2')``.

For example, if you use the sale module, you can configure an action to see all the products that have been sold for a given selection of sale orders.

### Configure

* Go to 'Setting / Technical / Actions / Server Actions'

* Create a new item

ir_actions_server_form

* Set a name that will be used for the Contextual Action

* Choose a model

* Select 'Navigate' option in the field 'Action To Do'

* Then, select a field of this model. It can be a ``one2many``, ``many2many``,  ``many2one`` field.

* You can then optionally select another fields that belong to the previous selected field relation model

* Once done, click on the 'Create Contextual Action' to generate a new 'More options' button.

**Extra options**

* You can define groups whose members will have access to that option.

* You can select an action that will be used to display the target field list.

### Usage

* Go to a tree view, for a model for which you have defined a navigation action.

* click on 'Action' Button

res_users_tree

* then click on the name of the configured Action and see the results

res_partner_category_tree


legalsylvain:
hi could you take a look on that migration PR ? 
CC : contributors  : @AshishHirapara (migration to V14)
CC original reviewers : @flotho, @hbrunn, @elvise, @francesco-ooops
 

Thanks !

francesco-ooops:
@legalsylvain rebase?

server_action_navigate/hooks.py
```diff
@@ -0,0 +1,7 @@
+from odoo import SUPERUSER_ID, api
+
+
+def uninstall_hook(cr, registry):
+    env = api.Environment(cr, SUPERUSER_ID, {})
+    env["ir.actions.server.navigate.line"].search([]).unlink()
+    env["ir.actions.server"].search([("state", "=", "navigate")]).unlink()
```
hbrunn, line 7: easy win re coverage to call this in some test?

hbrunn (approved):
I'd suggest to also add a test that edits an action with a tests.common.Form to have the domain computation covered

legalsylvain:
@hbrunn : done. Thanks for your review.

PR #277, opened 2024-03-22
[FIX] base_user_role: Hide role fields from non-admin users.

amh-mw:
This avoids an access error when a non-admin user attempts to view a list of users.

Steps to Reproduce:

1. --init=base_user_role,crm
2. Login as admin/admin
3. Create six test users
4. Create test role
5. Assign test role to some of the test users
6. Logout
7. Login as demo/demo
8. Edit a contact
9. On the Sales and Purchase tab, click into Salesperson and Search More...

FAIL: Access Error: You are not allowed to access 'User role' (res.users.role) records. This operation is allowed for the following groups: - Administration/Access Rights Contact your administrator to request access if necessary.

### DIFF
--- a/base_user_role/models/user.py
+++ b/base_user_role/models/user.py
@@ -11,12 +11,14 @@ class ResUsers(models.Model):
         inverse_name="user_id",
         string="Role lines",
         default=lambda self: self._default_role_lines(),
+        groups="base.group_erp_manager",
     )
     role_ids = fields.One2many(
         comodel_name="res.users.role",
         string="Roles",
         compute="_compute_role_ids",
         compute_sudo=True,
+        groups="base.group_erp_manager",
     )
 
     @api.model


legalsylvain (changes requested):
Hi @amh-mw. Thanks for reporting. 
agree with the bug you are raising. 
however, i think res.user.role and res.users.role.line should follow the same access right as res.groups in odoo core. 
Set the same read right for group_user should fix the problem.


see ir.model.access.csv : 
```
"access_res_groups_group_erp_manager","res_groups group_erp_manager","model_res_groups","group_erp_manager",1,1,1,1
"access_res_groups_group_user","res_groups group_user","model_res_groups",group_user,1,0,0,0

```

amh-mw:
> however, i think res.user.role and res.users.role.line should follow the same access right as res.groups in odoo core. Set the same read right for group_user should fix the problem.

With my security hat on, I don't see any reason that the average user has any business knowing what roles other users have.

legalsylvain:
> With my security hat on, I don't see any reason that the average user has any business knowing what roles other users have.

But for the time being, he has the access to the all groups for all the users. (see odoo core definition). So it is not very consistent to give access to all the groups, but hide role. Don't you think ?

amh-mw:
> But for the time being, he has the access to the all groups for all the users. (see odoo core definition). So it is not very consistent to give access to all the groups, but hide role. Don't you think ?

Two wrongs don't make a right. 😉

legalsylvain:
> Two wrongs don't make a right. 😉

yeah ! ;-) indeed ! Well, My point is more : is Odoo core bad ? 
-> If yes, we should make a PR to fix it. (or to know why they choose that implementation). 
-> If no, we have to follow the same convention.

What do you think ?

amh-mw:
> What do you think ?

I think bombs away. Draft pull request: https://github.com/odoo/odoo/pull/158915

PR #278, opened 2024-03-26
[FIX] Secure Odoo multi company, by only showing roles for non-compan…

rconjour:
We noted that when multiple companies are selected, you will have all the roles, of those companies combined. For that reason, i extended the code that if there are multiple companies selected, you only have the roles that are not linked to any specific company.

### DIFF
--- a/base_user_role_company/models/user.py
+++ b/base_user_role_company/models/user.py
@@ -2,6 +2,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 from odoo import api, models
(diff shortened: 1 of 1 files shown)


amh-mw (changes requested):
What if the user has the same role for each company?

rconjour:
> What if the user has the same role for each company?

@amh-mw if you configure that correctly, you'll leave the "company" field empty, and then this role is applied on every company, like this:
image

The role "dev" applies to all companies the user has access to, but sales manager, purchasing and ceo only to one specific defiened company.

PR #282, opened 2024-04-04
[FIX] base_user_role: Set group on access_wizard_groups_into_role

hwangh95:
From 17.0 onwards, any access right rule with set permissions must have a group on it, otherwise it will throw a warning on install

### DIFF
--- a/base_user_role/__manifest__.py
+++ b/base_user_role/__manifest__.py
@@ -4,7 +4,7 @@
 
 {
     "name": "User roles",
-    "version": "17.0.1.0.1",
+    "version": "17.0.1.0.2",
     "category": "Tools",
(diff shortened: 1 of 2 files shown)


amh-mw (approved):

legalsylvain (approved):
make sense.

base_user_role/__manifest__.py
```diff
@@ -4,7 +4,7 @@
 
 {
     "name": "User roles",
-    "version": "17.0.1.0.1",
+    "version": "17.0.1.0.2",
```
legalsylvain, line 7: For your information, this change is not necessary, as done by the bot when merging.

PR #286, opened 2024-04-23
[ADD][15.0] base_ical: Module to expose iCalendar over URL

fkantelberg:
This PR continues #231 and has implemented the following:

- Usage of existing apikeys of Odoo instead of a new model with tokens
- Offers an advanced mode which helps with more complex iCalendar files (allows to use the defined Odoo functions `_get_ics_file` to generate events)
- Add additional help in the form view


fkantelberg:
FYI @hbrunn @amh-mw 

FYI @jans23 In the advanced mode using `calendar = record._get_ics_file()` for `calendar.event` should also include VALARM.

base_ical/controllers/main.py
```diff
@@ -0,0 +1,54 @@
...
+    @http.route(
+        "/base_ical/<int:calendar_id>",
+        auth="none",
+        csrf=False,
+        methods=["GET"],
+        type="http",
+    )
+    def get_ical(self, calendar_id, access_token=None):
+        if not access_token:
+            raise request.not_found()
```
amh-mw: 401 Unauthorized

base_ical/controllers/main.py
```diff
@@ -0,0 +1,54 @@
...
+
+        user_id = request.env["res.users.apikeys"]._check_credentials(
+            scope=f"odoo.plugin.ical.{calendar_id}",
+            key=access_token,
+        )
+        if not user_id:
+            raise request.not_found()
```
amh-mw: 401 Unauthorized

base_ical/models/base_ical_token.py
```diff
@@ -0,0 +1,22 @@
+# Copyright 2023 Hunki Enterprises BV
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl-3.0)
+
+
+import secrets
+
+from odoo import fields, models
+
+
+class BaseIcalToken(models.Model):
```
amh-mw: Delete unused file

base_ical/controllers/main.py
```diff
@@ -0,0 +1,54 @@
...
+
+        calendar = (
+            http.request.env["base.ical"]
+            .sudo()
+            .search([("id", "=", calendar_id)])
```
amh-mw: Replace `search(...)` with `browse(calendar_id)` and relevant exception handling?

base_ical/controllers/main.py
```diff
@@ -0,0 +1,54 @@
...
+        user_id = request.env["res.users.apikeys"]._check_credentials(
+            scope=f"odoo.plugin.ical.{calendar_id}",
+            key=access_token,
+        )
+        if not user_id:
+            raise request.not_found()
+
+        calendar = (
+            http.request.env["base.ical"]
+            .sudo()
```
amh-mw: Given that users have read access to base.ical model per the security csv, replace `sudo()` with `with_user(user_id)` from below?

base_ical/controllers/main.py
```diff
@@ -0,0 +1,54 @@
...
+            .search([("id", "=", calendar_id)])
+            .with_user(user_id)
+        )
+        if not calendar:
+            return request.not_found()
+
+        records = calendar._get_items()
```
amh-mw, line 43: My suggestion above means that this call to `_get_items()` is no longer in sudo() mode and could return fewer results. I'm open to an argument on this.

base_ical/__manifest__.py
```diff
@@ -0,0 +1,28 @@
+# Copyright 2023 Hunki Enterprises BV
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl-3.0)
+
+{
+    "name": "Readonly publishing of calendars",
+    "summary": "Provide (readonly) .ics URLs to calendar-like models",
+    "version": "15.0.1.0.0",
+    "development_status": "Alpha",
+    "category": "Tools",
+    "website": "https://github.com/OCA/server-backend",
```
amh-mw, line 10: Should this land in https://github.com/OCA/calendar instead?

amh-mw (changes requested):

base_ical/controllers/main.py
fkantelberg: There is a big difference. `browse` doesn't check against existing (you would have to use `exists` afterwards which also results in a query. Also `search` would automatically hide inactive ones. I favor search here.

base_ical/__manifest__.py
fkantelberg: Not sure how we want to sort it. You can publish leaves => holidays. You can publish activities as TODO => social(?). You can publish a calendar => calendar. You can also publish attendance => attendance. server-backend might fit as general place.

You will probably use it in your calendar application but I'm not sure if this is the correct criteria.

base_ical/controllers/main.py
fkantelberg: Before is wasn't su too. because as @hbrunn mention `with_user` sets the su flag to false per default. Either way I guess that without sudo would be correct because you don't want to allow access to stuff a user can't see. If you link it with a nextcloud you should do it with an user with proper rights.

hbrunn: I agree it should be as it is - .with_user but no sudo

hbrunn (approved):

amh-mw (approved):
I still think this would land better in OCA/calendar (especially with regard to adding vobject to requirements.txt), but otherwise have no complaints.

PR #291, opened 2024-06-04
[17.0][MIG] base_user_role_history: Migration to 17.0

MohamedOsman7:


CRogos (changes requested):
When adding/removing a role to a user, the history entry is created.
When adding/removing a user from a role, no entry is created. (same bug in v16)
image

CRogos (approved):

gbrito (approved):

CRogos:
@pedrobaeza could you merge?

PR #294, opened 2024-06-13
[16.0][ADD] group_user_management

remi-filament:
This module adds a new res_groups for User Management as a solution for the issue that Admin Rights res_groups gives too much rights to users, allowing them to elevate their rights to Administration / Settings.

See also https://github.com/odoo/odoo/issues/119772#issuecomment-2140636367 for the position on Odoo that Administration / Access Rigths is considered by Odoo an administrator user not willing to see all menus from Administration / Settings.


group_user_management/models/res_groups.py
```diff
@@ -0,0 +1,70 @@
...
+
+from lxml import etree
+from lxml.builder import E
+
+from odoo import api, models
+
+from odoo.addons.base.models.ir_model import MODULE_UNINSTALL_FLAG
+
+
+class GroupsView(models.Model):
```
legalsylvain:
```suggestion
class ResGroups(models.Model):
```

group_user_management/models/res_partner.py
```diff
@@ -0,0 +1,24 @@
...
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    signup_token = fields.Char(
+        copy=False,
+        groups="base.group_erp_manager, group_user_management.group_user_manager",
+        compute="_compute_token",
+        inverse="_inverse_token",
+    )
```
legalsylvain:
```suggestion
    signup_token = fields.Char(
        groups="base.group_erp_manager, group_user_management.group_user_manager",
    )
```

when overloading fields, only redefine changed values.

group_user_management/models/res_groups.py
```diff
@@ -0,0 +1,70 @@
...
+    _inherit = "res.groups"
+
+    # The following lines are copied / modified from Odoo base module,
+    # res_users.py file, same function _update_user_groups_view(self)
+    @api.model
+    def _update_user_groups_view(self):
```
legalsylvain, line 15: Could you flag the lines that are modified ? I'm not sure to understand Why you have to overload this function. Could you elaborate ? 

thanks !

legalsylvain (commented):
some remarks inline. Thanks for sharing your work !

group_user_management/models/res_groups.py
remi-filament: Thanks @legalsylvain  for looking into it.
I kept it as is since Odoo names the class that way here : https://github.com/odoo/odoo/blob/66e7c919c57d5495aaf7705b641bed2070e1adf8/odoo/addons/base/models/res_users.py#L1367 (which is the same file where they already have Groups class defined) and also here : https://github.com/odoo/odoo/blob/66e7c919c57d5495aaf7705b641bed2070e1adf8/addons/account/models/res_users.py#L30
But fine with me I can rename that ResGroups

group_user_management/models/res_groups.py
remi-filament: I will add these in comments, for now, lines 25-27, 35-40, 62-70 are copied from Odoo the other not (specific view and specifix xpath).
Let me try to explain, because for now it does not work on CI tests since base_portal_type also overrides this method.

Odoo is creating in base module a dummy view here : https://github.com/odoo/odoo/blob/66e7c919c57d5495aaf7705b641bed2070e1adf8/odoo/addons/base/views/res_users_views.xml#L349-L357 with XML-id = base.user_groups_view
This view is then overriden in Python code in function _update_user_groups_view() (https://github.com/odoo/odoo/blob/66e7c919c57d5495aaf7705b641bed2070e1adf8/odoo/addons/base/models/res_users.py#L1402)
Basically, it loops over all res_groups category and for each it would create a group and corresponding access rights in res_users form view (either as boolean or as selection).
You get at the end something like : 
```xml
    <group string="Administration">
      <newline/>
      <field name="sel_groups_2_3" attrs="{'readonly': [('sel_groups_1_9_10', '!=', 1)]}" on_change="1"/>
      <newline/>
    </group>
```

I want to hide this Administration <group> for any user which does not have group_erp_manager, so I looked for the following options : 

1. Update the function that returns the categories of res_groups so that Administration category is not returned (this is what is done in account module to hide the groups group_account_readonly and group_account_user here https://github.com/odoo/odoo/blob/66e7c919c57d5495aaf7705b641bed2070e1adf8/addons/account/models/res_users.py#L34). The drawback of this option is that noone gets to see those in res_users form view, not event users with group_erp_manager or group_system), which is not ideal
2. The other option is to inherit the view with the list of groups in order to display this Administration <group> only for users with group_erp_manager. In order to do that, the only option I see is to update Python code (because that group view is regenerated everytime a new group is created for instance). Looking at how the Python code creating that view was made, I did not see a way to update it properly only where I needed to without monkey patching / rewriting everything, so I looked for another option : creating yet another view which would be overriden by the same Python code after the first one is created.

group_user_management/models/res_groups.py
legalsylvain: Yes. Odoo doesnt have strong convention on tha topic. OCA does. 
Better to follow OCA rules

group_user_management/models/res_groups.py
remi-filament: I had a lot of issues when updating module that Odoo would not find groups fields and after tryting many solutions, I simplified the way it is done so that I do not have to copy code from Odoo (and thus reverted license to AGPL).
I did the same as done in https://github.com/OCA/server-backend/blob/16.0/base_portal_type/models/res_groups.py : update view after it has been updated by Odoo main function.

The main drawback of this solution is that view will be updated every time it enters the above Odoo function since it is modified by extra module so the condition here : https://github.com/odoo/odoo/blob/66e7c919c57d5495aaf7705b641bed2070e1adf8/odoo/addons/base/models/res_users.py#L1525 will always be True...

PR #297, opened 2024-07-15
[16.0] [MIG] base_ical: migration to 16.0

hbrunn:
/ocabot migration base_ical


base_ical/views/res_users.xml
```diff
@@ -0,0 +1,47 @@
...
+                            <button
+                                type="object"
+                                name="action_new_url"
+                                string="New URL"
+                            />
+                            <field name="name" />
+                        </tree>
+                    </field>
+
+                    <label for="api_key_ids" string="Existing Access" />
```
hbrunn, line 24: @fkantelberg you left a note here to reevaluate this for v16 - what did you have in mind? I find it helpful to have the field here too, so that users don't have to manage their calendar urls via the api keys tab.

We might consider adding a field that actually filters for the scope though, as the domain on the view doesn't do that

fkantelberg: In 15.0 there was a `groups="base.group_no_one"` which limits the visbility more than in 16.0 where this was removed. (developer api keys => api keys) This is the reason I added the following code in 15.0. In 16.0 you already see the api keys properly. Adding a new field would show them on 2 spots which might also confuse people.

I would remove it and just go with the standard here. The name of the API key should be feasable.

neoneddy:
Anything we can do to get this pulled in officially?

jcdrubay:
> Anything we can do to get this pulled in officially?

Hi @neoneddy 
You could help by doing a [1] code review
Or doing [2] functional test on runboat (as soon as the build is complete). Normally, the build should have complete for a long time, so maybe a Maintainers needs to re-trigger the build.
Or [3] ensure that the README reflects well what the module is doing (you can add your comments in the tab Files Changed of this PR)

Once you have done one of the three tasks, then you can add submit a code review from the tab Files Changed, and then on the top right you can use the button "Review Changes" to comment or approve.

Honeyxilia:
This PR is interesting to us, and we'd like to review it towards a potential fusion - could you reopen this PR ? ty

PR #153, opened 2022-03-10
[14.0][FW][FIX] base_import_match: importing related o2m records

dreispt:
Forward port of https://github.com/OCA/server-backend/pull/142

PR #299, opened 2024-07-17
[14.0][FIX] base_import_match: importing related o2m records

sergiocorato:
The original code did improperly handle importing
related one2many records where at best all their
attribute values would get overwritten by last
of them. This was caused by mapping them only by
their first part.


sergiocorato:
Replaces #153

PR #300, opened 2024-07-29
[15.0][FIX] base_user_role: prevent crash when using 'res.users' (x2m) fields in forms

IT-Ideas:
Prior to this commit, when `res.users` is added to a model as being a x2m field, the `advanced search` is causing an error.

### DIFF
--- a/base_user_role/models/user.py
+++ b/base_user_role/models/user.py
@@ -11,12 +11,14 @@ class ResUsers(models.Model):
         inverse_name="user_id",
         string="Role lines",
         default=lambda self: self._default_role_lines(),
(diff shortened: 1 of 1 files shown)


sebalix (approved):

IT-Ideas:
Hey @sebalix, is there any fw port? As this will occur in next versions too.

sebalix:
@IT-Ideas I'll port it to next releases yes

EDIT: ported to 16.0 and 17.0, thank you!

amh-mw:
Fixes https://github.com/OCA/server-backend/pull/277

PR #305, opened 2024-08-21
[15.0][FIX] change role_line_ids default value to override create

PaoloYam:
Populating role_line_ids this way doesn't work in this case. The previous method indeed return the right value but at creation, role_line_ids is empty in vals resulting of a new user having no role attributed even though the default_user has one. 

Feature already has test. Test added for coverage.

### DIFF
--- a/base_user_role/models/user.py
+++ b/base_user_role/models/user.py
@@ -10,7 +10,6 @@ class ResUsers(models.Model):
         comodel_name="res.users.role.line",
         inverse_name="user_id",
         string="Role lines",
-        default=lambda self: self._default_role_lines(),
         groups="base.group_erp_manager",
     )
     role_ids = fields.One2many(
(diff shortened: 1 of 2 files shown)


santostelmo (approved):

base_user_role/models/user.py
```diff
@@ -21,21 +20,23 @@ class ResUsers(models.Model):
...
         default_user = self.env.ref("base.default_user", raise_if_not_found=False)
-        default_values = []
-        if default_user:
-            for role_line in default_user.with_context(active_test=False).role_line_ids:
-                default_values.append(
+        role_line_model = self.env["res.users.role.line"]
+        if not default_user:
+            return
+        role_lines = default_user.with_context(active_test=False).role_line_ids
+        for user in record:
```
santostelmo, line 29: Not sure this is the most correct way todo it. But it does the job

victorvermot (approved):

sebalix (changes requested):
I am not sure to understand the change.
From a UX point of view, when I create a user (through the form view) I expect to see the role lines pre-populated thanks to the `default` attribute of the field, I can change them and then I expect these values to be stored as well when saving the form.

With the current changes we are losing this feature, roles will be populated when saving the form.

If your goal is to populate the roles when calling directly `create`, you can rework the code without losing the existing feature. Also, you should not force the default user roles if `vals` has them already set, only populate them if they were not set.
```python
@api.model
def create(self, vals):
    if "role_line_ids" not in vals:
        # Populate roles from default user
    return super().create(vals)
```

Test cases:
- `create({})` => no role defined, we populate roles from the default user
- `create({'role_line_ids': False})` => we do not want to set any role on purpose, we do nothing
- `create({'role_line_ids': [(6, 0, [IDs...]))}` => roles have been defined with explicit values, same as previous example we do not add default user roles on top of that

PR #308, opened 2024-09-05
[14.0] base_user_role: warning when adding admin to empty role

ajaniszewska-dev:
When creating a new user role, it is possible to add a user to that role without having added any groups. If this is saved, all that user’s access rights will be wiped and he will not able to login anymore (Internal Server Error). It can cause problems when happens for admin user.

### DIFF
--- a/base_user_role/models/user.py
+++ b/base_user_role/models/user.py
@@ -1,6 +1,6 @@
 # Copyright 2014 ABF OSIELL <http://osiell.com>
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
-from odoo import api, fields, models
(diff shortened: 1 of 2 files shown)


ivantodorovich (approved):

base_user_role/models/user.py
```diff
@@ -19,6 +19,19 @@ class ResUsers(models.Model):
         compute_sudo=True,
     )
 
+    @api.onchange("role_line_ids")
```
sebalix, line 22: I was thinking of a computed boolean field to display or not a `<div class="alert alert-warning" role="alert">` in the forms, less intrusive than a popup. What do you think?

PR #304, opened 2024-07-31
[16.0][FW] base_user_role: port PR #250 and #262

sebalix:
Port the following PRs from 15.0 to 16.0:
- https://github.com/OCA/server-backend/pull/262 (to ease the port I preferred to backport this feature from 17.0 as it was already ported there in a squashed commit in the [migration PR](https://github.com/OCA/server-backend/pull/253/commits))
- https://github.com/OCA/server-backend/pull/250

PR #309, opened 2024-09-05
[16.0][FIX] base_user_role: Improve test validation sorting values

victoralmau:
Improve test validation sorting values

Related to https://github.com/OCA/server-backend/pull/304

Please @pedrobaeza and @sebalix can you review it?

@Tecnativa

### DIFF
--- a/base_user_role/tests/test_user_role.py
+++ b/base_user_role/tests/test_user_role.py
@@ -269,4 +269,4 @@ class TestUserRole(TransactionCase):
         new_role = self.role_model.search([("name", "=", "Test Role")])
         self.assertTrue(new_role)
         # Check that the role has the correct groups
-        self.assertEqual(new_role.implied_ids.ids, user_group_ids)
+        self.assertEqual(set(new_role.implied_ids.ids), set(user_group_ids))


pedrobaeza (commented):
I think there's no need to sort. Sets already compare equality without taking order into account.

victoralmau:
I have proposed this change because it solves this error (installing several core modules). Sorted is something that is already being used in `other base_user_role` tests.

```
FAIL: TestUserRole.test_group_groups_into_role
  File "/opt/odoo/auto/addons/base_user_role/tests/test_user_role.py", line 272, in test_group_groups_into_role
    self.assertEqual(new_role.implied_ids.ids, user_group_ids)
AssertionError: Lists differ: [12, 8, 80, 77, 28, 82, 32, 35, 17, 19, 63, 53, 3[121 chars], 78] != [12, 8, 38, 82, 80, 32, 77, 17, 19, 63, 28, 53, 3[121 chars], 78]
First differing element 2:
80
38
```

base_user_role/tests/test_user_role.py
```diff
@@ -269,4 +269,6 @@ def test_group_groups_into_role(self):
         new_role = self.role_model.search([("name", "=", "Test Role")])
         self.assertTrue(new_role)
         # Check that the role has the correct groups
-        self.assertEqual(new_role.implied_ids.ids, user_group_ids)
+        self.assertEqual(
+            sorted(set(new_role.implied_ids.ids)), sorted(set(user_group_ids))
+        )
```
pedrobaeza: I mean, please try this:

```suggestion
        self.assertEqual(set(new_role.implied_ids.ids), set(user_group_ids))
```

victoralmau: It also works!

pedrobaeza (approved):

CarlosRoca13 (approved):

PR #311, opened 2024-10-04
[ADD] gitlab module

ByteMeAsap:
This module does not provide a functionality by itself but a base model to implement gitlab related features


gitlab/README.rst
```diff
@@ -0,0 +1,85 @@
...
+    :target: https://translation.odoo-community.org/projects/server-backend-16-0/server-backend-16-0-gitlab
+    :alt: Translate me on Weblate
+.. |badge5| image:: https://img.shields.io/badge/runboat-Try%20me-875A7B.png
+    :target: https://runboat.odoo-community.org/builds?repo=OCA/server-backend&target_branch=16.0
+    :alt: Try me on Runboat
+
+|badge1| |badge2| |badge3| |badge4| |badge5|
+
+This module does not provide a functionality by itself but a base model
+to implement gitlab related features
```
legalsylvain, line 32: this looks very generic, and quite hard to understand the perimeter of this module.
Could you mention links to extra modules that depends on this module ? 

thanks !

ByteMeAsap: @legalsylvain ,the extra module is not yet added as its dependent on this module.The module is aimed at recording user activities like commits, pull requests (merge requests), opening issues etc.

legalsylvain: Hi maybe we could share concept and code with https://github.com/OCA/interface-github module, dont you think ? 
With some refactor it looks me possible at a first sight

ByteMeAsap: Here is the link to the PR for the extra module that depends on this module - https://github.com/OCA/vertical-association/pull/178. Actually, its not identical to the interface-github

PR #317, opened 2024-10-28
[18.0][ADD] polars_process: process polars dataframe from file

bealdav:


bealdav:
see https://github.com/OCA/server-backend/pull/320

PR #318, opened 2024-10-28
[18.0][ADD] polars_db_process: get db query in polars dataframe

bealdav:


bealdav:
see https://github.com/OCA/server-backend/pull/320

PR #319, opened 2024-10-30
[16.0][FIX] base_user_role: default roles

danielduqma:
When granting access to a contact, a copy from *Portal User Template* is made for it. But applied default roles are the ones from *Default User Template*, causing an error when trying to assign both internal and portal types to new user.

## Steps to reproduce

1. Create a new role, assigning it group *User types / Internal User*
2. Assign this role to user *Default User Template*
3. Grant access to a *res.partner* (*Actions > Grant portal access* and then *GRANT ACCESS*)

image

FL-556-4543

### DIFF
--- a/base_user_role/models/user.py
+++ b/base_user_role/models/user.py
@@ -34,15 +34,21 @@ class ResUsers(models.Model):
         default_user = self.env.ref("base.default_user", raise_if_not_found=False)
         default_values = []
         if default_user:
-            for role_line in default_user.with_context(active_test=False).role_line_ids:
-                default_values.append(
-                    {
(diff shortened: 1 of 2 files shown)


miguelcb2003 (approved):
LGTM!

StefanRijnhart (commented):
Looks like a good fix! Can you maybe reproduce the error in a test to guarantee your fix will keep on working in the future?

danielduqma:
Done @StefanRijnhart, thanks for the feedback!

Regards

StefanRijnhart (approved):
Thanks for adding the test!

dreispt (approved):

PR #322, opened 2024-11-20
[FIX][15.0] base_ical: User creation and repetition rules

fkantelberg:
This PR wants to use 2 problems:
- Problem when creating multiple users at once with an automatic calendar throws an exception when allowing the calendar
- Repetition rule in the ICS files lead to duplication for calendar.events because Odoo already created multiple records out of it and this leads to duplicated entries

### DIFF
--- a/base_ical/models/base_ical.py
+++ b/base_ical/models/base_ical.py
@@ -39,6 +39,10 @@ class BaseIcal(models.Model):
     )
     preview = fields.Text(compute="_compute_preview")
     code = fields.Text()
+    remove_repetition_rule = fields.Boolean(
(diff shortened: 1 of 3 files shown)


hbrunn (approved):

base_ical/models/base_ical.py
```diff
@@ -194,6 +199,9 @@ def _get_items(self, limit=None):
 
     def _get_ical(self, records=None, limit=None):
         """Return the vcalendar as text"""
+        if not self.model_id:
```
hbrunn, line 202: you could mark this branch as `# pragma: nocover`

PR #324, opened 2024-12-15
[17.0][FIX] base_user_role_company: wrong arguments in new odoo releases

ramiadavid:
In new releases of odoo17 a new argument has been added to the `web_load_menus` function

https://github.com/odoo/odoo/blob/f360eb2f008cdc9de88f3db058dd24085b5be499/addons/web/controllers/home.py#L66


ramiadavid:
@dreispt @rconjour 

Please, can you review and approve this? Odoo does not work when installing the module if it is not fixed

dreispt (approved):

PR #315, opened 2024-10-23
[18.0][MIG] base_user_role_company: Migration to 18.0

BertVGroenendael:

PR #328, opened 2025-01-08
[18.0][MIG] base_user_role_company: Migration to 18.0

yankinmax:
Thanks to @BertVGroenendael the `base_user_role_company` is migrated here: https://github.com/OCA/server-backend/pull/315
This PR is opened to fast-line the merge process.

@StefanRijnhart @ristecona


vvrossem (approved):

StephaneMangin (approved):
LGTM

base_user_role_company/models/role.py
```diff
@@ -0,0 +1,42 @@
...
+    @api.constrains("user_id", "company_id")
+    def _check_company(self):
+        for record in self:
+            if (
+                record.company_id
+                and record.company_id != record.user_id.company_id
+                and record.company_id not in record.user_id.company_ids
+            ):
+                raise ValidationError(
+                    _(
```
Ricardoalso: _"You can replace the call to translate a string (_) with self.env._ for getting some performance improvement in some cases. See https://github.com/odoo/odoo/pull/174844."_
```suggestion
                    self.env._(
```

base_user_role_company/views/role.xml
```diff
@@ -0,0 +1,20 @@
...
+        <field name="inherit_id" ref="base_user_role.view_res_users_form_inherit" />
+        <field name="arch" type="xml">
+            <field name="role_id" position="attributes">
+                <attribute name="domain">[]</attribute>
+            </field>
+            <xpath
+                expr="//field[@name='role_line_ids']//field[@name='role_id']"
+                position="after"
+            >
+                <field name="allowed_company_ids" invisible="1" />
```
Ricardoalso:
```suggestion
                <field name="allowed_company_ids"  />
```

StefanRijnhart: For my information, why is this field visible now?

Ricardoalso: Oh my bad  ! my comment was misleading and targeted the invisible parameter instead of the whole line. 

I was referencing the migration [guideline](https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-18.0) below : 

_Missing fields used in domains and other attributes are added to the view as invisible automatically. If you encounter fields in views with invisible="True", they can probably be removed. Reference: https://github.com/odoo/odoo/pull/137031_

StefanRijnhart: Agreed, the field can be removed!

yankinmax: Rather a good catch! Missed that!

StefanRijnhart: Sorry, I didn't test this before but I'm testing it now on runboat and it seems that `allowed_company_ids` in the domain is evaluated as the companies that are enabled for the current user. In the screenshots you can see that the user 'stefan' only has access to the main company. The admin user has both companies enabled. In the role configuration, I can select both companies, instead of just the company that is enabled for the edited user.

Can you check if this is caused by the requested change?

image
image
image

base_user_role_company/models/ir_http.py
```diff
@@ -0,0 +1,22 @@
...
+    def session_info(self):
+        """
+        Based on the selected companies (cids),
+        calculate the roles to enable.
+        A role should be enabled only when it applies to all selected companies.
+        """
+        result = super().session_info()
+        if self.env.user.role_line_ids:
+            cids_str = request.httprequest.cookies.get("cids", str(self.env.company.id))
+            cids = [int(cid) for cid in cids_str.split(",")]
```
StefanRijnhart: It seems cids are split by dashes in the cookie as of https://github.com/odoo/odoo/commit/e8d80ceb348#diff-747c1e41e030ec3b083ce9f5405cd3a0f950656ad6a59dbfc39e744955b81472.
If I enable more than one company on a user with roles, I get
```
  File "/mnt/data/odoo-addons-dir/base_user_role_company/models/ir_http.py", line 20, in session_info
    cids = [int(cid) for cid in cids_str.split(",")]
  File "/mnt/data/odoo-addons-dir/base_user_role_company/models/ir_http.py", line 20, in <listcomp>
    cids = [int(cid) for cid in cids_str.split(",")]
ValueError: invalid literal for int() with base 10: '2-1'
```

StefanRijnhart (changes requested):

yankinmax:
@StefanRijnhart I need to debug the issue and will get with an update soon

base_user_role_company/models/ir_http.py
yankinmax: Yep, agree with you we should split companies' ids by dashes.

StefanRijnhart:
@yankinmax Wait, you just overwrote most of the migration changes: https://github.com/OCA/server-backend/compare/159adb6ffde762fde388818051c4b42de4cf2618..0d5896d724568725e66730da1842b3b34ef782ca?diff=unified&w=#diff-37a501b2f98480be68553ef3a061c6f92bc31c331e2a7ccf59a9eafb6bb542de

yankinmax:
> @yankinmax Wait, you just overwrote most of the migration changes: https://github.com/OCA/server-backend/compare/159adb6ffde762fde388818051c4b42de4cf2618..0d5896d724568725e66730da1842b3b34ef782ca?diff=unified&w=#diff-37a501b2f98480be68553ef3a061c6f92bc31c331e2a7ccf59a9eafb6bb542de

My bad, I'll update now

base_user_role_company/models/ir_http.py
StefanRijnhart: Thanks, that seems fixed now.

StefanRijnhart:
Maybe I don't know how to configure this module, but on runbot my session crashes if I log in as a user with different roles in two companies, and both companies enabled. At that point, Odoo redirects to portal url `my`, then crashes on currency_id access.

image
image
image


```
  File "/opt/odoo/odoo/http.py", line 2366, in __call__
    response = request._serve_db()
  File "/opt/odoo/odoo/http.py", line 1894, in _serve_db
    return self._transactioning(
  File "/opt/odoo/odoo/http.py", line 1966, in _transactioning
    exc.error_response = self.registry['ir.http']._handle_error(exc)
  File "/opt/odoo/addons/http_routing/models/ir_http.py", line 610, in _handle_error
    code, html = 418, request.env['ir.ui.view']._render_template('http_routing.http_error', values)
  File "/opt/odoo/odoo/addons/base/models/ir_ui_view.py", line 2191, in _render_template
    return self.env['ir.qweb']._render(template, values)
  File "/opt/odoo/odoo/tools/profiler.py", line 306, in _tracked_method_render
    return method_render(self, template, values, **options)
  File "/opt/odoo/odoo/addons/base/models/ir_qweb.py", line 601, in _render
    result = ''.join(rendering)
  File "<265>", line 67, in template_265
  File "<265>", line 56, in template_265_content
  File "<190>", line 668, in template_190
  File "<190>", line 428, in template_190_content
  File "<190>", line 40, in template_190_t_nocache_1
odoo.addons.base.models.ir_qweb.QWebException: Error while render the template
AccessError: You are not allowed to access 'Currency' (res.currency) records.

This operation is allowed for the following groups:
	- Administration/Settings
	- Invoicing/Administrator
	- User types/Internal User
	- User types/Portal
	- User types/Public

Contact your administrator to request access if necessary.
Template: web.frontend_layout
Path: /t/html/head/script[2]/t
Node: <t t-out="json.dumps(request.env[\'ir.http\'].get_frontend_session_info())"/>
2025-01-24 09:29:26,187 18 INFO bc4602eac-8098-4d0e-9ef0-7753ccdae140 werkzeug: 65.108.98.234 - - [24/Jan/2025 09:29:26] "GET /my HTTP/1.1" 500 - 16 0.004 0.009
```
It's not a migration issue because something similar happens in v17, but I think this module should be allowed in the next version before it's fixed (unless someone can point out a beginner mistake I'm making with the configuration).

base_user_role_company/views/role.xml
StefanRijnhart: I see on the v17 runbot that this is a migration issue as it does not occur there. Maybe it can be fixed by leaving in the `allowed_user_ids` field (but with `column_invisible`)

yankinmax: I'll check, thanks

yankinmax:
hello @StefanRijnhart , this migration (runboat) was tested by the Camptocamp business analyst.
"The Readme.srt specifies that “A company specific Role will only be enabled if it is set for all the currently selected companies.”. 

So if a user has a role as Sales Admin on company A and a role as Purchase Admin on company B, then if both companies are selected no role is enabled. 
Same behavior if a user has a role as Sales Admin for company A and a role as Sales Own Document for company B,  if both companies are selected then no role is enabled. 

Now, if a user has a role as Sales Admin for company A and a role as Sales Own Document for company A and company B, then, if both companies are selected, only the role Sales Own Document will be applied."

Can you pls check again and in case it's ok we go further here?

base_user_role_company/views/role.xml
StefanRijnhart: Solved, thanks for the update.

StefanRijnhart (approved):
Thanks for the clarification! I think we can rely on the functional analyst in that case.

Issue #327, opened 2025-01-07
[17.0] base_role_user : failing tests on using OCA CI 


bouktaibMohssin:
## Module
base_role_user

## Describe the bug

There is an issue with the ordering of lists in the base_role_user module that is causing a test failure. The error occurs when the test test_group_groups_into_role checks if the implied_ids.ids from a new role matches the user_group_ids. Although the lists contain the same elements, they are in a different order, which leads to an AssertionError.
![image](https://github.com/user-attachments/assets/f14ff9c6-b5af-4477-81cf-2c5d54544feb)

## To Reproduce
**Affected versions**: 17

1. Run the tests using the OCA CI for the base_role_user module.
2. Compare the two lists that are expected to be identical.
Example: 
   The lists are as follows:
    
    List 1 (from new_role.implied_ids.ids): [8, 28, 35, 60, 23, 45, 16, 26, 27, 9, 1, 13, 46, ... , 21]
    List 2 (from user_group_ids): [8, 28, 23, 45, 35, 60, 16, 26, 27, 9, 1, 13, 46, ... , 21]
    The first differing element is at index 2:
    
    List 1 has 35 at index 2, while List 2 has 23 at that position.

**Expected behavior**
The two lists (new_role.implied_ids.ids and user_group_ids) should be identical, with the elements appearing in the same order.

**Additional context**
test function that cause where the error comes from 
https://github.com/OCA/server-backend/blob/77db565b2537b7e15df80795b6d706fdcb1916ea/base_user_role/tests/test_user_role.py#L272

Issue #321, opened 2024-11-04
[16.0] base_user_role: test cases are failed if we use this module together with auditlog


AungKoKoLin1997:
Is there anyone who is using `base_user_role` with [auditlog](https://github.com/OCA/server-tools/tree/16.0/auditlog)?
I just found if we put these two together in the same repos, the tests CI is failed and the test cases from base_user_role are failing.

@qrtl

ng-ife:
@AungKoKoLin1997 Can you ass the test that actually fails? I'm facing a similar issue on a customer system (but with auditlog not installed).
Could be related.

PR #330, opened 2025-01-09
[17.0][FIX] base_user_role : fix tests to group groups into a role

mb-ife:


ng-ife:
Add force group recalculation in role assignment tests to handle
group synchronization issues that occur when base_user_role and
auditlog modules are both installed. Without forcing updates, tests
fail in odoo.sh and OCA CI environments due to group assignment
mismatches.

Force parameter in set_groups_from_roles() ensures proper group
synchronization regardless of auditlog's presence. No issues occur
when base_user_role runs standalone.

Fix applied to all role assignment test methods to maintain consistent
behavior across test scenarios.

ng-ife (approved):

ng-ife:
Hello @pedrobaeza 
Can you review this PR? Thank you

pedrobaeza:
I can't say about this one.

ng-ife:
@pedrobaeza Do you know how to ping? You made the "mistake" to help me once at last OCA days but I'm more then willing to spread my pings.

pedrobaeza:
You can ping module maintainers. Formally, the PR shouldn't change the module version (it's done by the bot), and that way, you clean it without the README changes.

sebalix:
There are fixes which have been done for other Odoo versions, and we should probably try to port them first (the use of `sorted` for instance has been fixed in 15.0 by using `set`). I'm trying to port some of the PRs done in 15.0 to 17.0 first:

- https://github.com/OCA/server-backend/pull/335

sebalix:
@mb-ife @ng-ife , can you check if the mentioned PR above fixed your issues? And/or adapt your PR if it remains fixes to apply

mb-ife:
Hello @sebalix ,
Thanks for the updates , we have issues mainly with testing we'll double check again and give you feedback.

sebalix:
I also ported missing commits for 16.0 and 18.0 there:
- https://github.com/OCA/server-backend/pull/336
- https://github.com/OCA/server-backend/pull/337

mb-ife:
Hello @sebalix 

After checking the new version it solve for us this issue here with the sorted groups https://github.com/OCA/server-backend/issues/327 , but still have an issue when combined with auditlog module https://github.com/OCA/server-backend/issues/321 , the current PR is currently solving this at the moment.

sebalix:
@mb-ife Ok, thanks for the feedback. What I do not like the way it is fixed here, is that it doesn't really test the feature as it should work: we update the roles on the user, and the call to `set_groups_from_roles` is implicit (we should not have to call it manually in tests) and we should get the expected results.

So it's like hidding the issue, without explaining why it behaves this way: we probably have something else to fix in the design (in `auditlog` or `base_user_role`, or if not, a comment explaining the reason why we are taking this shortcut).

StefanRijnhart:
Indeed, if this is an issue in combination with auditlog, it's the issue that needs fixing, not the test. Can you tell us how to reproduce the issue?

mb-ife:
Hello @sebalix, @StefanRijnhart 

After investigating the test failure, I realized that simply adding a manual call to set_groups_from_roles(force=True) is not an adequate solution to the underlying issue. When the auditlog and base_user_role modules are installed together, the automatic group assignment for roles does not work as expected.

To reproduce we only need to run a test with base_user_role and auditlog on the same folder. This is happening on Odoo 16 as well, and you can see a similar issue here https://github.com/OCA/server-backend/issues/321. The error is triggered on OCA CI as well as on Odoo.sh.

The easiest approach seems to be forcing the groups update in base_user_role. I would appreciate a second opinion on this approach.

PR #334, opened 2025-01-27
[18.0][IMP] base_user_role - add path in act_window to fit with new url logic

jdidderen:
Simple improvement to have a proper url 

Before :
image

After : 
image

### DIFF
--- a/base_user_role/views/role.xml
+++ b/base_user_role/views/role.xml
@@ -85,10 +85,11 @@
             </search>
         </field>
     </record>
-    <record model="ir.actions.act_window" id="action_res_users_role_tree">
+    <record id="action_res_users_role_tree" model="ir.actions.act_window">
         <field name="name">Roles</field>
-        <field name="type">ir.actions.act_window</field>
         <field name="res_model">res.users.role</field>
+        <field name="path">user-role</field>
+        <field name="view_mode">list,form</field>
         <field name="view_id" ref="view_res_users_role_tree" />
     </record>
     <menuitem


base_user_role/views/role.xml
```diff
@@ -85,10 +85,11 @@
             </search>
         </field>
     </record>
-    <record model="ir.actions.act_window" id="action_res_users_role_tree">
+    <record id="action_res_users_role_tree" model="ir.actions.act_window">
```
legalsylvain, line 88: Why this change ?

jdidderen-nsi: For the sake of consistency. All other records in the file begin with id=“” and end with model=“”.
If it bothers you, I can remove it.

jdidderen: For the sake of consistency. All other records in the file begin with id=“” and end with model=“”.
If it bothers you, I can remove it.

base_user_role/views/role.xml
```diff
@@ -85,10 +85,11 @@
...
         <field name="name">Roles</field>
-        <field name="type">ir.actions.act_window</field>
         <field name="res_model">res.users.role</field>
+        <field name="path">role</field>
```
StefanRijnhart: Ambiguity can quickly arise using these paths, I fear. Can you make this `user-role`, in line with the model's `_description`?

jdidderen: Done 👍

StefanRijnhart (approved):
Thanks!

legalsylvain (approved):
/ocabot merge patch

PR #338, opened 2025-01-28
[18.0][MIG] base_user_role_history: Migration to 18.0.

mcodes-o:
Migrated to 18.0


mcodes-o:
@ThomasBinsfeld Since you're listed as a maintainer, do you have time to look at this PR? Thanks!

ThomasBinsfeld (approved):

JessBrandl (approved):
Code LGMT, tested in runboat and works as intended

base_user_role_history/models/res_users.py
```diff
@@ -0,0 +1,30 @@
...
+            res = self.env["base.user.role.line.history"].search(
+                [("user_id", "=", user.id)], limit=1, order="id desc"
+            )
+            user.last_role_line_modification = res.create_date if res else False
+
+    def show_role_lines_history(self):  # pragma: no cover
+        self.ensure_one()
+        domain = [("user_id", "=", self.id)]
+        return {
+            "name": _("Roles history"),
```
CRogos: You can replace the call to translate a string ( _ ) with self.env._ for getting some performance improvement in some cases. See https://github.com/odoo/odoo/pull/174844.

CRogos (commented):
LGTM, but some small findings:

Squash administrative commits (if any) with the previous commit for reducing commit noise. Check https://github.com/OCA/maintainer-tools/wiki/Merge-commits-in-pull-requests#mergesquash-the-commits-generated-by-bots-or-weblate for details.

image

base_user_role_history/models/res_users.py
mcodes-o: @CRogos Thanks, has been fixed in commit f59c8100a844610371c3dbd20fd586d264ba2f48

mcodes-o:
> LGTM, but some small findings:
> 
> Squash administrative commits (if any) with the previous commit for reducing commit noise. Check https://github.com/OCA/maintainer-tools/wiki/Merge-commits-in-pull-requests#mergesquash-the-commits-generated-by-bots-or-weblate for details.
> 

@CRogos  I've squashed the commits and pushed. Thanks for the review!

base_user_role_history/models/res_users.py
CRogos: Can you squash this together with the "Migration to 18.0" commit?

CRogos:
@mcodes-o these need to be squashed too:
image

base_user_role_history/tests/test_base_user_role_history.py
```diff
@@ -0,0 +1,142 @@
+# Copyright 2019 ACSONE SA/NV
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).µ
+
+from datetime import date, timedelta
+
+from odoo.tests.common import TransactionCase
+
+
+# DEPRECATED: This tests are deprecated but stay to show that the new code is working.
+class TestBaseUserRoleHistory(TransactionCase):
```
CRogos: Use BaseCommon as base test class to reduce overhead from tracking and boost the test suite.

Maybe also check if the depricated tests can be removed?

CRogos (commented):
@mcodes-o could you merge the last two commits?

CRogos:
@mcodes-o are you still working on this?

base_user_role_history/tests/test_base_user_role_history.py
mcodes-o: @CRogos , Removed the deprecated tests, since it duplicated the functionality. 

The existing test had a small error in it, fixed that.  The test wrote the date as a string, and then did an update using a string again. But by then, everything is converted to date objects. Fixed the test by writing date objects both times, the expected format.. Code now works as it should do..

base_user_role_history/tests/test_base_user_role_history.py
```diff
@@ -1,142 +0,0 @@
-# Copyright 2019 ACSONE SA/NV
```
etobella: Why did you remove the tests???

base_user_role_history/tests/test_res_users_role.py
```diff
@@ -0,0 +1,181 @@
...
+
+        self.assertEqual(len(history_lines_all), 2)
+
+        history_lines_add = history_lines_all.filtered(
+            lambda x: x.performed_action == "edit"
+        )
+        self.assertEqual(len(history_lines_add), 1)
+
+        history_line_1 = history_lines_add[0]
+        self.assertEqual(history_line_1.performed_action, "edit")
```
CRogos, line 73: Maybe we could add another assert which I found in the deleted tests. The rest I agree.
```suggestion
        self.assertEqual(history_line_1.performed_action, "edit")
        self.assertEqual(history_line_1.old_role_id, role_01)
        self.assertEqual(history_line_1.new_role_id, role_01)
```

base_user_role_history/tests/test_base_user_role_history.py
rousseldenis: Yes, please restore original file and apply needed changes on that one.

mcodes-o: The original file was deleted because it was marked as deprecated, and all the functionality is tested in the file test_res_users_role.py. Code coverage confirms that all the paths are still touched. @CRogos found one extra assertion we need to do to keep the tests the same.

No point in testing the same thing twice.

etobella: @mcodes-o the tests had a lot of explanations and comments that were interesting.  Please restore them.

The currents test might have the same coverage, but sometimes it is not just the test, because sometimes it includes a lot of information of how to use it and the reason behind it.

CRogos: In other words the tests are not self explaining because they are to complicated 😅 ...

I personally don't like to many actions in one test case. Each test should test one case/action and assert the result. In case of an error, you don't need to analyse which part of the test has failed.

Doing many steps in one test, like in `def test_write_role_lines_on_user(self)` is a bad practice in my opinion.

If it is necessary to test adding Roles to User and User to Roles I don't know. But if you insist to restore the test file, we should at least remove the depricated. (and maybe split test_write_role_lines_on_user)

CRogos: @mcodes-o I am not convinced, but they do not really harm. Therefore could you restore the file so that we can continue?

CRogos:
@mcodes-o are you still working on this?

CRogos:
to bring this PR to an end, I've re-added the test file and created a new PR: https://github.com/OCA/server-backend/pull/376

@rousseldenis @etobella could you do a review

rousseldenis (approved):

etobella:
Superseeded by #376

PR #340, opened 2025-02-05
[FIX] base_user_role_company: Do not auto install this module.

legalsylvain:
the _company module is an extra module used in a multi company context (and only in some situation). It is totally useless in mono company. As 2 modules exists, it doesn't makes sense to make the second auto_installable.


legalsylvain:
@dreispt : could you take a look on this trivial one ? 

Thanks !

rvalyi (approved):

TomVermeerenDynapps (approved):

PR #344, opened 2025-03-04
[MIG] server_action_navigate: Migration to 17.0

Daemo00:
Standard migration from `16.0` (https://github.com/OCA/server-backend/tree/79523b003b0989b18a88991f2af86e3bbcc7b67d/server_action_navigate).


Daemo00:
@legalsylvain since you have been working on this, would you like to have a look?

server_action_navigate/views/view_ir_actions_server.xml
```diff
@@ -2,43 +2,45 @@
...
-                        <field name="navigate_model_technical_name" invisible="1" />
-                        <field
-                            name="navigate_action_id"
-                            attrs="{'invisible': [('navigate_line_ids', '=', [])]}"
-                        />
-                    </group>
-                </page>
-            </xpath>
+            <notebook position="after">
+                <notebook invisible="state != 'navigate'">
```
legalsylvain: Why créate a notebook with a single page, if there is no notebook in the base module ?

legalsylvain (approved):
Minor remarks inline. 
Otherwise, lgtm. Thanks !

server_action_navigate/views/view_ir_actions_server.xml
Daemo00: Thanks to you for having a look!

Right, it seemed more aligned to the "Code" action and to how the module used to work in `16.0`, but maybe in `17.0` it makes little sense; now it's a `group`.
Please check if now it looks better

server_action_navigate/views/view_ir_actions_server.xml
```diff
@@ -2,43 +2,37 @@
...
-                        <field name="navigate_model_id" />
-                        <field name="navigate_model_technical_name" invisible="1" />
-                        <field
-                            name="navigate_action_id"
-                            attrs="{'invisible': [('navigate_line_ids', '=', [])]}"
-                        />
-                    </group>
-                </page>
-            </xpath>
+            <notebook position="after">
```
legalsylvain, line 13: Why not add a page, if it exist in core module ?

Sorry but i really dont understand the changes you introduces in that file.

legalsylvain (changes requested):

server_action_navigate/views/view_ir_actions_server.xml
Daemo00: Don't worry, I understand, you probably haven't seen the news in the parent view: https://github.com/odoo/odoo/blob/ef6379a4bcf3cddbb0f261da69ff460ec11c8b21/odoo/addons/base/views/ir_actions_views.xml#L411 (it mainly changed in https://github.com/odoo/odoo/commit/7be6d6e3214e3d5be2b3085ec4b3b8cc2fd228e2)
It looks like now every kind of action shows their own `div`/`field`/`notebook`/whatever.

The only existing `notebook` is the one that is only displayed when the action is of type `code`: in that case a notebook with two pages is shown.
That is why my first approach was to copy it and simply add a new `notebook` with the same page that was in `16.0` for this module.

But then your first review https://github.com/OCA/server-backend/pull/344#discussion_r2072661208 highlighted that a notebook with a single page is not much of a notebook, so I extracted the existing `group`, and here we are.

Of course I could edit the `invisible` property of the existing notebook and hide the existing pages and only show the one of this module when the server action is `navigate` but that would be a mess, don't you think?

legalsylvain: Ok ! Indeed. Hard to overload this view with such core design. 

Thanks for all the explanation !

legalsylvain (approved):
/ocabot merge nobump

PR #347, opened 2025-03-26
[18.0][FIX] base_user_role_company: Roles are filtered so none which exceed date_to are used

TomVermeerenDynapps:
Fixed a bug where disabled roles (because past date_to) were used.

### DIFF
--- a/base_user_role_company/tests/test_use_only_enabled_roles.py
+++ b/base_user_role_company/tests/test_use_only_enabled_roles.py
@@ -1,0 +1,62 @@
+# Copyright 2021 Open Source Integrators
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+from datetime import timedelta
+
+from odoo import fields
+from odoo.tests.common import TransactionCase
+
+
+class TestUserRoleCompany(TransactionCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.company1 = cls.env.ref("base.main_company")
+        # GROUPS for roles
+        cls.groupA = cls.env.ref("base.group_user")
+        cls.groupB = cls.env.ref("base.group_system")
+        # ROLES
+        cls.Role = cls.env["res.users.role"]
+        cls.roleA = cls.Role.create({"name": "Role internal user"})
+        cls.roleA.implied_ids |= cls.groupA
+        cls.roleB = cls.Role.create({"name": "Role system user"})
+        # USER with roles
+        cls.roleB.implied_ids |= cls.groupB
+        cls.User = cls.env["res.users"]
+        user_vals = {
+            "name": "Role test user",
+            "login": "role_test_user",
+            "company_ids": [(fields.Command.set([cls.company1.id]))],
+            "role_line_ids": [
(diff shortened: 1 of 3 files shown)


StefanRijnhart (commented):
Maybe use this convenience method: https://github.com/OCA/server-backend/blob/14.0/base_user_role/models/user.py#L54-L55
Can you add a small test?

TomVermeerenDynapps:
@StefanRijnhart _get_enabled_roles is the method I am overwriting, so instead of calling it (recursion error) again I will use res, this  has the same functionality. I also added test scenarios.

base_user_role_company/tests/test_use_only_enabled_roles.py
```diff
@@ -0,0 +1,61 @@
+# Copyright 2021 Open Source Integrators
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+from datetime import timedelta
+
+from odoo import fields
+from odoo.tests.common import TransactionCase
+
+
+class TestUserRoleCompany(TransactionCase):
+    def setUp(self):
```
StefanRijnhart: Please refactor to use 
```
@classmethod
def setUpClass(cls)
```
so that it will only have run once to be used for both tests.

base_user_role_company/tests/test_use_only_enabled_roles.py
```diff
@@ -0,0 +1,61 @@
...
+        self.roleA = self.Role.create({"name": "Role internal user"})
+        self.roleA.implied_ids |= self.groupA
+        self.roleB = self.Role.create({"name": "Role system user"})
+        # USER with roles
+        self.roleB.implied_ids |= self.groupB
+        self.User = self.env["res.users"]
+        user_vals = {
+            "name": "Role test user",
+            "login": "role_test_user",
+            "company_ids": [(6, 0, [self.company1.id])],
```
StefanRijnhart:
```suggestion
            "company_ids": [fields.Command.set([self.company1.id])],
```

base_user_role_company/tests/test_use_only_enabled_roles.py
```diff
@@ -0,0 +1,61 @@
...
+            "role_line_ids": [
+                (0, 0, {"role_id": self.roleA.id}),
```
StefanRijnhart:
```suggestion
                fields.Command.create({"role_id": self.roleA.id}),
```
also below

StefanRijnhart (commented):
Great thanks! Only some minor comments now.

StefanRijnhart (commented):
Thanks for the update! Please squash the commits.

TomVermeerenDynapps:
Comments resolved & commits squashed! Thank you for review!

StefanRijnhart (commented):
Apologies, one thing I missed: the module name needs to be featured in the commit message, so `[FIX] base_user_role_company: ...` (like the PR title actually).

TomVermeerenDynapps:
Should be resolved now:) thank you!

StefanRijnhart (approved):
Thanks for the quick responses!

benwillig (approved):

dreispt (approved):

PR #349, opened 2025-04-15
[ADD][16.0] base_user_group_mgmt: manage groups/roles assignation with validation

benwillig:
This PR creates two modules which allow users to assign group and roles to a user only if a workflow is approved.

* base_user_group_mgmt: add/remove groups from users, add/remove groups from groups
* base_user_group_mgmt_role: add/remove users from roles, add/removes groups from roles

Related fields are set readonly (only in the UI)

PR #351, opened 2025-04-23
[ADD][16.0] base_group_erp_user & base_group_erp_user_role

benwillig:
Theses new modules adds a new first level in "Administration" groups to allow create users, groups and roles without being able to change the groups of a user, the groups of a role, the groups of a role and the roles assigned to users.


legalsylvain:
It looks similar to something I developped, some monthes ago. However, I don't see the possibility to avoid right escalation. Or did I missed something ? 

https://github.com/grap/grap-odoo-incubator/tree/16.0/user_limited_access_settings

benwillig:
@legalsylvain It looks a bit similar indeed, about rights escalation, we use theses modules with the ones in this PR https://github.com/OCA/server-backend/pull/349 so there is only a group of users who can edit user rights

sbejaoui (approved):
code review + Functional test

base_group_erp_user/models/res_users.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+class ResUsers(models.Model):
+    _name = "res.users"
+    _inherit = ["res.users", "mixin.erp.user.forbidden.fields"]
+
+    @api.model
+    def _get_erp_user_system_forbidden_fields(self):
+        return [
+            "groups_id",
```
acsonefho, line 18: You have to call the super and `extends`/`append` the result with your new field

base_group_erp_user/models/res_groups.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+class ResGroups(models.Model):
+    _name = "res.groups"
+    _inherit = ["res.groups", "mixin.erp.user.forbidden.fields"]
+
+    @api.model
+    def _get_erp_user_system_forbidden_fields(self):
+        return [
+            "implied_ids",
```
acsonefho, line 18: You have to call the super here too

acsonefho (changes requested):

base_group_erp_user/models/res_users.py
benwillig: I should have raised a NotImplementedError in the mixin to be honest so I don't think it's necessary here

PR #352, opened 2025-04-23
[15.0][ADD] model_access_restriction

Tisho99:
This module adds a new model to configure Odoo permissions, the "Model
Access Restrictions"

This model allows to restrict the access to a model for all users except
the ones that belong to at least one group of a list of allowed groups.

While Odoo's default access rules provide permissions, and having one
already gives you access, these new rules remove them, and failing to
comply with one restricts your access.

T-7772


sara-castello (approved):
Functional review LGTM

model_access_restriction/models/ir_model_access_restriction.py
```diff
@@ -0,0 +1,119 @@
...
+        string="Allowed Groups",
+        comodel_name="res.groups",
+        relation="model_access_restriction_group_rel",
+        column1="model_access_restriction_id",
+        column2="group_id",
+        ondelete="restrict",
+    )
+    perm_read = fields.Boolean(string="Apply for Read", default=True)
+    perm_write = fields.Boolean(string="Apply for Write", default=True)
+    perm_create = fields.Boolean(string="Apply for Create")
```
manuelregidor: Could you explain why this field has a different default value?

model_access_restriction/models/ir_model_access_restriction.py
```diff
@@ -0,0 +1,119 @@
...
+    perm_unlink = fields.Boolean(string="Apply for Delete", default=True)
+
+    def init(self):
+        self.env.cr.execute(
+            """
+            CREATE INDEX IF NOT EXISTS ir_model_access_restriction_model_name_id_idx
+            ON ir_model_access_restriction(model_id,name,id)
```
manuelregidor, line 51: Do you need to add name to index?

model_access_restriction/tests/test_model_access_restriction.py
```diff
@@ -0,0 +1,95 @@
...
+from odoo.exceptions import AccessError, ValidationError
+from odoo.tests.common import TransactionCase
+
+
+class TestModelAccessRestriction(TransactionCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.model = cls.env.ref("base.model_res_partner")
+        cls.model2 = cls.env.ref("base.model_res_partner_category")
```
manuelregidor, line 13: Do you need model2?

model_access_restriction/models/ir_model_access_restriction.py
Tisho99: Because restricring the read permission to all users is too agressive

model_access_restriction/models/ir_model_access_restriction.py
Tisho99: Yes, it was requested to be a index of the sort fields

model_access_restriction/tests/test_model_access_restriction.py
Tisho99: I have changed the tests to use this model

manuelregidor (approved):
Techinical reviw. LGTM

Tisho99:
@ValentinVinagre @HaraldPanten

ValentinVinagre:
@luis-ron , can you check that everything is working in the client test environment? @manuelregidor  will remain idle until this check is made.

ValentinVinagre:
@pedrobaeza @etobella what do u think?

ValentinVinagre:
> @pedrobaeza @etobella what do u think?

Do you find the module interesting?

pedrobaeza:
IMO, this module will be a source of un/noticed access errors due to the strict nature of the configuration, and also it's a bomb in the hands of a child if the users have administrator permissions.

PR #354, opened 2025-04-28
[18.0][MIG] base_user_effective_permissions: Migration to 18.0

BhaveshHeliconia:


BhaveshHeliconia:
Hey @OCA/tools-maintainers would be great if someone could have a look at this module.

legalsylvain:
@hbrunn : could you take a look on this migration ?

base_user_effective_permissions/models/res_users_effective_permission.py
```diff
@@ -49,21 +49,17 @@ def _generate_permissions(self, user):
...
-                    )
-                    for operation in operations
-                }
-            )
-            vals.update(
-                {
-                    f"{operation}_domain": IrRule._compute_domain(
+            for operation in operations:
+                try:
+                    model.check_access(operation)
```
hbrunn: this is not the equivalent of v16's `model.check_access_rights` - use `ir.model.access#check` instead

base_user_effective_permissions/models/res_users_effective_permission.py
```diff
@@ -49,21 +49,17 @@ def _generate_permissions(self, user):
...
+                    vals[f"{operation}_permission"] = True
+                except Exception:
+                    vals[f"{operation}_permission"] = False
+                try:
```
hbrunn: under which conditions is this expected to raise? I'd expect something to be very wrong when this happens, and then we shouldn't squelch the exception

hbrunn (changes requested):
/ocabot migration base_user_effective_permissions

BhaveshHeliconia:
@hbrunn : I’ve made the changes as per your suggestion. Could you please review? Thanks!”

base_user_effective_permissions/models/res_users_effective_permission.py
hbrunn: you didn't answer this question

BhaveshHeliconia: The _compute_domain() method can raise an AccessError when:

1. The user doesn't have access to certain ir.rule records
2. Domain evaluation fails due to missing fields or invalid domain syntax
3. Security restrictions prevent domain computation for specific models

In these cases, setting an empty domain "[]" is appropriate as it indicates no domain-based restrictions beyond the basic access rights check already performed by ir.model.access#check.
However, you're right that we shouldn't squelch all exceptions. I'll update the code to catch only AccessError

hbrunn: obvious copy&paste from coding assistant, no interest to continue discussion

BhaveshHeliconia: @hbrunn here I accept that I used AI but its only to clean up the wording, since sometimes I struggle to words things well. Sorry if that caused any confusion.
Next onwards I will try to best to answer well.

base_user_effective_permissions/=
```diff
@@ -0,0 +1,22 @@
+# Copyright 2023 Hunki Enterprises BV
```
yvaucher: Please remove this file.

It seems to be a duplicate of the __manifest__ file. It was left over in v16, I created a PR here: #425

yvaucher (changes requested):

base_user_effective_permissions/=
BhaveshHeliconia: Thanks for pointing it out. I've removed the file.

base_user_effective_permissions/models/res_users_effective_permission.py
thomaspaulb: @hbrunn From my side I would say that this comment is addressed by catching only AccessError?

hbrunn: it's nonsense to catch any error here

hbrunn: to better mimic what Odoo does this could however not call _compute_domain at all if vals[f"{operation}_permission"] is False, and do all of this in the same loop anyways

hbrunn (changes requested):
thanks for sanitizing the loop. remaining issue is that the domain widget crashes on some permission records, I suspect you need to filter out cases where `model._abstract` is `True`

BhaveshHeliconia:
> thanks for sanitizing the loop. remaining issue is that the domain widget crashes on some permission records, I suspect you need to filter out cases where `model._abstract` is `True`

Thanks for pointing this out. Please have another look when you get a chance.

hbrunn (approved):

yvaucher (approved):

PR #364, opened 2025-06-13
[18.0][IMP] base_user_role: Add role_new_test_user helper function.

amh-mw:
This allows developers to easily create test users with given roles.

### DIFF
--- a/base_user_role/tests/__init__.py
+++ b/base_user_role/tests/__init__.py
@@ -1,1 +1,2 @@
+from . import common
 from . import test_user_role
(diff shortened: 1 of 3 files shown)


etobella (approved):
It would be nice to use it inside the tests, because you probably need it there somewhere :smile:

amh-mw:
> It would be nice to use it inside the tests, because you probably need it there somewhere 😄

😅 You're absolutely right. I use this function extensively in our private modules and it slipped my mind to add tests for it here.

StefanRijnhart (approved):

Issue #232, opened 2023-07-21
[16.0] base_portal_type: indicator for portal groups as a field/flag on res.groups instead of category_id


BenediktZimmer4C:
Hi,
(I am fully aware that the module base_portal_type has only been commited a few days ago and is in alpha.)

Did you consider making the indicator whether a security group is an option for portal users as a boolean flag on the res.groups model?
At the moment - as far as I understand the description and the code - groups need to be put into the special category.
In earlier versions of Odoo there was a field "is_portal" on res.groups, which was used for displaying the groups in the "grant portal access" wizard on res.partner. Something similar could be applied.

Generally speaking, I am very much willing to help developing this module - if there are any open issues.
We are currently in the process of migrating our Odoo environment with target version 16 and I was very surprised noticing the feature disappear along the way to grant portal access on a more granular level - so we are basically required to either use this particular module or a implement solution inhouse.
I am not familiar with the exact process here on github (sorry!), but if you create an issue labeled "help wanted", I will try to contribute.

PR #368, opened 2025-06-29
[16.0][IMP] base_portal_type: use boolean field to define portal-group

goreki:
This PR adjusts the circumstances under which a res.group becomes available for portal users:
In addition to the magic module category in the data.xml of this module, a boolean field 'portal' can be set on a group to make it selectable for portal users.
For backwards compatibility I have kept the module category, though.

This makes it also possible to put portal-groups into actual categories and then have them displayed in those categories on the res.users form view.

It addresses an old/stale/closed issue: https://github.com/OCA/server-backend/issues/232

Not sure if actively pinging the main contributor is best practice or frowned upon, but since this module was initiated by you, maybe you want to take a look :smile: 
@hbrunn

### DIFF
--- a/base_portal_type/models/res_groups.py
+++ b/base_portal_type/models/res_groups.py
@@ -1,16 +1,17 @@
 # Copyright 2023 Hunki Enterprises BV
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl-3.0)
 
-
 from lxml import etree
 
-from odoo import api, models
+from odoo import api, fields, models
 
 from odoo.addons.base.models.res_users import name_boolean_group, name_selection_groups
 
 
 class ResGroups(models.Model):
     _inherit = "res.groups"
+
+    portal = fields.Boolean("Is Portal Group")
 
     @api.model
     def _update_user_groups_view(self):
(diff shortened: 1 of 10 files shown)


base_portal_type/models/res_groups.py
```diff
@@ -24,51 +26,71 @@ def _update_user_groups_view(self):
...
+            [
+                "|",
+                (
+                    "category_id",
+                    "=",
+                    self.env.ref("base_portal_type.category_portal_type").id,
+                ),
+                ("portal", "=", True),
+            ]
+        )
```
hbrunn, line 56:
```suggestion
        ).with_context(lang=None)
```

be sure that you use English for names

base_portal_type/models/res_groups.py
```diff
@@ -24,51 +26,71 @@ def _update_user_groups_view(self):
...
+
+        arch_root = arch.xpath("//field[@name='groups_id']")[0]
+        if len(arch_root) and portal_groups:
```
hbrunn:
```suggestion
        if not portal_groups:
            return result
        for arch_root in arch.xpath("//field[@name='groups_id']"):
```

no need to do the write below if we don't change anything, and `arch.xpath("//field[@name='groups_id']")[0]` raises if the xpath is empty for whatever reason

base_portal_type/models/res_groups.py
```diff
@@ -24,51 +26,71 @@ def _update_user_groups_view(self):
...
+                ),
+            )
+            for app_id, app_name in portal_groups.mapped("category_id").name_get() + [
+                (False, "Other")
+            ]:
+                portal_groups_in_app = portal_groups.filtered(
+                    lambda r: (r.category_id.id or False) == app_id
+                )
+                if not portal_groups_in_app:
+                    continue
```
hbrunn, line 85: you can test this by adding a test where you delete the other groups. while being on it, you can also add a test for the case where there are no portal groups at all

hbrunn (changes requested):
please add a sentence or two about the new flag in the readme, also explaining in which cases you'd want the flag. when done, squash your commits

goreki:
> please add a sentence or two about the new flag in the readme, also explaining in which cases you'd want the flag. when done, squash your commits

thank you for all the feedback - I will integrate it as suggested during the weekend. Would you be okay with me adding myself as a contributor in the manifest?

When this PR is going to be merged, I will also take a look at migration to v17 and 18 (I see there is already a PR, which obviously does not have my adjustments included yet). I am not worried about any big changes - more about following the guidelines, so this will be another learning experience for me.

hbrunn:
contributors go usually to https://github.com/goreki/server-backend/blob/16.0/base_portal_type/readme/CONTRIBUTORS.rst

base_portal_type/models/res_groups.py
goreki: true. I did it differently and moved the write() section into the if-block.
I think I now also understood that the whole function is called twice during install/upgrade - and the first time with the 'install_filename' in the context, which creates a placeholder view in the super() function, which therefore leaves us with the empty xpath result.
you didn't encounter that problem beforehand, because you where for-looping over the xpath results --> not doing a single loop in that case.

I didnt try, but I would even go as far as saying that our whole function _update_user_groups_view() can simply be skipped at the very top by checking if the 'install_filename' is in the context

base_portal_type/models/res_groups.py
goreki: done.
since the groups defined in the demo data can be adjusted and deleted by a user/developer, the tests could fail afterwards although the code is fine. So I also adjusted the existing test to not rely on the demo data

hbrunn (approved):
thanks

PR #372, opened 2025-07-30
[18.0][FIX] base_user_role: don't reset notification_type

jhemono:
The self-writable Notification field on users, which allows choosing to be notified by email or via push, actually sets or unsets a group on the user. This would conflict with role-based management of groups: when setting push notification, the field would be set and push notification received but the group would immediately be reset by base_user_role (and some UI element would not be shown) and, in some case like module activation, the notification type would be reset to mail, the field being recomputed from the user's groups.

This change excludes the group in question form being unset or set by roles. Exclusions can be added by overriding a method.

### DIFF
--- a/base_user_role/models/user.py
+++ b/base_user_role/models/user.py
@@ -64,6 +64,13 @@ class ResUsers(models.Model):
     def _get_enabled_roles(self):
         return self.role_line_ids.filtered(lambda rec: rec.is_enabled)
 
+    @api.model
+    def _get_self_writable_groups(self):
+        group = self.env.ref(
+            "mail.group_mail_notification_type_inbox", raise_if_not_found=False
+        )
+        return group or self.env["res.groups"]
+
     def set_groups_from_roles(self, force=False):
         """Set (replace) the groups following the roles defined on users.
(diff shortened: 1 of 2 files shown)


jhemono:
I didn't add unit tests because I don't know how to do so without depending on mail to get the excluded group. I'm open to inputs !

StefanRijnhart (commented):
This is a really important fix, thank you for that! I've encountered this issue multiple times now. I reckon you can test this with a post-install test, because mail would normally be installed at that point.

StefanRijnhart (approved):

espo-tony (approved):

amh-mw:
> I didn't add unit tests because I don't know how to do so without depending on mail to get the excluded group.

You could add an additional module `test_base_user_role` that depends on `base_user_role` and `mail`, then put the tests there.

StefanRijnhart:
@amh-mw A test is now in place.

jhemono:
Sorry. I've forgot to mention that I've added a test when I force-pushed. I solved the problem by skipping the test if mail is not installed but your solution would do it too.

PR #374, opened 2025-08-06
 [14.0][ADD] company_id field on ResUsersRoleLine

mathieudelva:


legalsylvain (changes requested):
Could you explain that change ?

PR #375, opened 2025-08-12
[18.0][FIX] base_user_role: fix field name ambiguity

StefanRijnhart:
Core Odoo defines a `Roles` field on res.partner, which is inherited on res.users.

This gives a recurring warning
```
Two fields (l10n_es_edi_facturae_ac_role_type_ids, role_ids) of res.users() have the same label: Roles. [Modules: None and base_user_role]
```

as well as real ambiguity in the list of views in search and export widgets.

https://github.com/odoo/odoo/blob/18.0/addons/l10n_es_edi_facturae/models/res_partner.py#L20-L21

Propose to change the field name to fix the ambiguity, and update labeling in other places for consistency.

### DIFF
--- a/base_user_role/models/res_groups.py
+++ b/base_user_role/models/res_groups.py
@@ -21,7 +21,7 @@ class ResGroups(models.Model):
     role_ids = fields.Many2many(
         comodel_name="res.users.role",
         relation="res_groups_implied_roles_rel",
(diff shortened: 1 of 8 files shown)


legalsylvain (approved):
Thanks !

/ocabot merge patch

PR #388, opened 2025-11-03
[18.0] [MIG] base_external_dbsource_mssql

Nikul-OSI:


dreispt:
@Nikul-OSI tests are failing

Nikul-OSI:
> @Nikul-OSI tests are failing

@dreispt It's not related to our code; it failed because the library didn't load properly.

`ImportError: libodbc.so.2: cannot open shared object file: No such file or directory
`

Nikul-OSI:
@max3903 @dreispt Tests and pre-commit now :green_heart:

base_external_dbsource_mssql/models/base_external_dbsource.py
```diff
@@ -0,0 +1,83 @@
...
+        ondelete={"mssql_alchemy": "cascade", "mssql_pyodbc": "cascade"},
+    )
+    PWD_STRING_MSSQL_ALCHEMY = "Password=%s;"
+    PWD_STRING_MSSQL_PYODBC = "Password=%s;"
+
+    def connection_close_mssql_alchemy(self, connection):
+        return connection.close()
+
+    def connection_close_mssql_pyodbc(self, connection):
+        return connection.close()
```
max3903, line 32: @Nikul-OSI Why duplicating the method if they do the same thing? Each method should handle the different library. Revert to what it was on sunday.

Nikul-OSI: @max3903 `connection_close` or `connection_open` function calls a subfunction `_get_adapter_method`, and this main function concatenates the adapter method with the Connector name.

For example, connection_close + mssql_pyodbc = `connection_close_mssql_pyodbc`

_get_adapter_method = https://github.com/OCA/server-backend/blob/18.0/base_external_dbsource/models/base_external_dbsource.py#L312

pedrobaeza:
Please solve conflicts.

Nikul-OSI:
> Please solve conflicts.

Done @pedrobaeza

christian-ramos-tecnativa:
I'm not entirely convinced about the approach taken for the MSSQL ODBC implementation. As it stands, the `open`, `close`, and `execute` methods are identical for both types, and the shared logic is relying on the `connector` field for filtering.

In my opinion, it would be cleaner to introduce a new field, `mssql_type` (defaulting to `alchemy`), in `base.external.dbsource`, and then use that field for filtering in the `_execute_mssql` method. This way, the rest of the logic can remain unchanged.
 @Nikul-OSI @max3903 WDYT

JordiMForgeFlow (approved):
functionally tested 👍🏼

pedrobaeza:
Superseded by #392

PR #390, opened 2025-11-18
[FIX] base_ical: fix res_users._compute_ical_ids

jans23:
The cal_ids field isn't writable by the user because it's not part of SELF_WRITEABLE_FIELDS hence the normal access check happens and a normal user isn't allowed to write to res.users (the exception only happens if all fields are part of SELF_WRITEABLE_FIELDS).

### DIFF
--- a/base_ical/models/res_users.py
+++ b/base_ical/models/res_users.py
@@ -12,7 +12,7 @@ class ResUsers(models.Model):
 
     def _compute_ical_ids(self):
         domain = [("allowed_users_ids", "=", self.env.uid)]
-        self.write({"ical_ids": self.env["base.ical"].search(domain)})
+        self.sudo().write({"ical_ids": self.env["base.ical"].search(domain)})
 
     @property
     def SELF_READABLE_FIELDS(self):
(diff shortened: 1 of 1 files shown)


base_ical/models/res_users.py
```diff
@@ -12,7 +12,7 @@ class ResUsers(models.Model):
 
     def _compute_ical_ids(self):
         domain = [("allowed_users_ids", "=", self.env.uid)]
-        self.write({"ical_ids": self.env["base.ical"].search(domain)})
```
hbrunn, line 15: @fkantelberg please remind me why this is a write and not an assignment?

and why is this not

```python
for this in self:
    domain = [("allowed_users_ids", "=", this.id)]
    this.ical_ids = self.env["base.ical"].search(domain)
```

? In the current form, this field should be wrong when some user looks at it for another user

@jans23 please test if the above fixes your problem too, as this seems more correct to me but I don't have a v14 instance ready to verify

jans23: Yes, it solves my issue.

BTW, it is version 15.0.

jans23: I changed this fix accordingly.

hbrunn (approved):

StefanRijnhart (approved):
/ocabot merge patch

PR #397, opened 2025-12-02
[18.0][ADD] base_user_role_activity: create a reminder activity before role assignment expires

CRogos:
Happy for any feedback and suggestions.


MohamedOsman7 (approved):
Code review LGTM

CRogos:
@pedrobaeza could you merge this?

pedrobaeza:
I don't use this one, sorry.

CRogos:
@jcdrubay, @novawish, @sebalix could you have a look on this base_user_role extension and add a review if you think this is usefull?

- Aggregates multiple expiring roles into one notification
- creates activity on res.partner not on expiring res.users.role.line
- notifies relevant manager
- activity automatically deleted, when role expiration is updated

CRogos:
@bofiltd could you also review this PR?

base_user_role_activity/models/user.py
```diff
@@ -0,0 +1,74 @@
...
+            employees = (
+                self.env["hr.employee"]
+                .sudo()
+                .search([("user_id", "=", user.id), ("parent_id", "!=", False)])
+            )
+            manager = employees.filtered(
+                lambda e: e.company_id == self.env.company
+            ).parent_id
+            if not manager and employees:
+                manager = employees[0].parent_id
```
sebalix, line 56: This module should not depend on `hr` IMO (and not on the CI of this repository neither).

Better to add a hook `_get_role_manager` that returns the current user by default, and create a glue module (in `OCA/hr`) that overrides it. What do you think?

sebalix (changes requested):
Great addition, however I have one blocking point.

base_user_role_activity/models/user.py
CRogos: I like the idea, but I think returning the current user by default is not correct. The method is called by a cron job and the activity should not be assigned to the bot user. 
But we could use self.id (the user who's role is about to expire) by default. 

The name of the glue module hr_user_role_activity?

PR #399, opened 2025-12-05
[19.0][MIG] base_user_role: rename user_ids to role_user_ids

JohanEndare:
## Summary
Migration of base_user_role module to Odoo 19.0 with field renaming to avoid conflicts.

## Changes
- Renamed `user_ids` field to `role_user_ids` to avoid naming conflicts
- Updated to use `fields.Command` for x2many operations (Odoo 19 best practices)
- Complete migration from 18.0 to 19.0

### DIFF
--- a/base_user_role/models/role.py
+++ b/base_user_role/models/role.py
@@ -24,8 +24,8 @@ class ResUsersRole(models.Model):
     line_ids = fields.One2many(
         comodel_name="res.users.role.line", inverse_name="role_id", string="Role lines"
     )
-    user_ids = fields.One2many(
-        comodel_name="res.users", string="Users list", compute="_compute_user_ids"
+    role_user_ids = fields.One2many(
+        comodel_name="res.users", string="Users list", compute="_compute_role_user_ids"
     )
     rule_ids = fields.Many2many(
         comodel_name="ir.rule",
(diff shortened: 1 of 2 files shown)


legalsylvain (changes requested):
Hi. could you elaborate where is the conflict ? 

thanks.

JohanEndare:
> Hi. could you elaborate where is the conflict ?
> 
> thanks.

Odoo logs a warning due to users being renamed to user_ids in Odoo 19. It has to do with the model inheritance

StefanRijnhart:
@JohanEndare There is no warning in the logs of this module when running tests (e.g. https://github.com/OCA/server-backend/actions/runs/21262764308/job/61194130829). What is the exact warning, and when does it occur?

CRogos:
@StefanRijnhart I assume @JohanEndare is referring to a change in res.group which is inherrited from.

There "users" changed to user_ids which now causes the conflict.

https://github.com/odoo/odoo/blob/f5da5197620ad12302b40ff1702a9cda5007aaa3/odoo/addons/base/models/res_groups.py#L17

amh-mw (changes requested):
1. Commit/pull request title should contain [FIX] not [MIG]
2. ~~Need a migration script or this is going to delete data~~ Edit: I temporarily forgot that fields.One2many isn't stored on model.

StefanRijnhart:
@CRogos Thanks for clarifying. The field name on our model masks the field name on the inherited model.

StefanRijnhart (approved):

PR #402, opened 2025-12-09
[FIX] base_user_role: Order res.users.role per name.

legalsylvain:
Rational, for the time being, there is no order defined. As a result, displayed are done by id, that doesn't make sense for end users


amh-mw (approved):
Code review only, LGTM.

PR #409, opened 2026-01-15
[19.0] [FIX] base_user_role: Fixes group view inheritance to show "User Roles" button

NachoAlesLopez:
In V18 and older, the group form view had a button to access which roles the group is applied per role:

image

image

This button isn't present on V19, currently:

image

This PR should fix this regression by changing the inheritance, as the newer Odoo version includes the required section to add buttons.

### DIFF
--- a/base_user_role/views/group.xml
+++ b/base_user_role/views/group.xml
@@ -5,23 +5,17 @@
         <field name="model">res.groups</field>
         <field name="inherit_id" ref="base.view_groups_form" />
         <field name="arch" type="xml">
-            <xpath expr="//sheet/group" position="before">
-                <div name="button_box" class="oe_button_box">
-                    <button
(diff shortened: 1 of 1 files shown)


amh-mw (approved):
Code review only, LGTM.

CRogos:
Screenshot LGTM and code also, but could not test because runboat is not available. Could you rebase?

Maybe also change the groups view back to a list like it was in v18?
image

StefanRijnhart (approved):
Thanks! Makes sense, given https://github.com/odoo/odoo/commit/968f524ffd666951c0edc6383188b988c5c3e1de#diff-1830e19d4bdede74bd1c6a48084739a33ab6c50ee5fc12bf73b14e61d5cd4193.

PR #413, opened 2026-01-21
[18.0][ADD] gitlab: Import gitlab data into Odoo

tarteo:


legalsylvain (changes requested):
hi. Thanks for your contribution. At first sight, I think that this module could better in another repo. (as per github modules that are here : https://github.com/OCA/interface-github)
a interface-gitlab module, or gitlab-connector ...

thanks !

tarteo:
Thanks for your reply, I was already wondering if this was the correct repo. I'll try to arrange a new o r different repo.

pedrobaeza:
Please both check https://github.com/OCA/interface-github/issues/137

PR #414, opened 2026-01-27
[16.0][IMP] Make base_external_system easily extendable

NL66278:
This module will be extended with modules to connect to http systems, with additional modules to use oauth authorization.

This code is running in production at several customer sites already, together with additional modules for which PR's will follow.


ntsirintanis (approved):

base_external_system/models/external_system.py
```diff
@@ -13,42 +14,47 @@ class ExternalSystem(models.Model):
...
-        help="This is the private key that is used for authenticating to "
-        "this system, if applicable.",
+        help="This is the private key that is used for authenticating to"
+        " this system, if applicable."
     )
     private_key_password = fields.Text(
-        help="This is the password to unlock the private key that was "
-        "provided for this sytem.",
+        help="This is the password to unlock the private key that was"
+        " provided for this sytem."
```
thomaspaulb: system.

base_external_system/models/external_system.py
```diff
@@ -13,42 +14,47 @@ class ExternalSystem(models.Model):
...
+    scheme = fields.Char(
+        help="This is the protocol used to communicate with the system."
     )
     host = fields.Char(
-        help="This is the domain or IP address that the system can be reached " "at.",
+        help="This is the domain or IP address that the system can be reached at."
     )
     port = fields.Integer(
-        help="This is the port number that the system is listening on.",
+        help="This is the port number that the system is listening on."
```
thomaspaulb: What generated these changes? If it was pre-commit, could you separate that in a separate commit so that it's clear which part is refactoring and which is pre-commit?

base_external_system/models/external_system.py
```diff
@@ -101,24 +111,55 @@ def client(self):
...
+            return True
+        except Exception as exc:
+            raise ValidationError(
+                _(
+                    "Unexpected error %(exception)s"
+                    " when connecting to %(system_name)s"
+                )
+                % {
+                    "exception": exc,
+                    "system_name": self.name,
```
thomaspaulb: Personally I always find it slightly annoying if exceptions are converted to something else than they previously were. At the very least can we log the traceback, either in the logfile or to the chatter if it exists? Then it is possible to know where things went wrong.

base_external_system/readme/CONFIGURE.md
```diff
@@ -0,0 +1,4 @@
+Configure external systems in Settings =\> Technical =\> External
+Systems
+
+[![Try me on Runbot](https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas)](https://runbot.odoo-community.org/runbot/149/10.0)
```
thomaspaulb: What does this link do here?

base_external_system/security/ir.model.access.csv
```diff
@@ -1,3 +1,2 @@
 id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
-access_external_system_os_admin,access_external_system_os_admin,model_external_system_os,base.group_system,1,1,1,1
```
thomaspaulb, line 2: Would this line need to be there for external_system_adapter_os still?

thomaspaulb (changes requested):
some minor comments

base_external_system/models/external_system.py
NL66278: Corrected

base_external_system/models/external_system.py
NL66278: Basically this is feedback to the user that he/she probably made an error configuring the connection. The action is to test the connection. So I think a ValidationError is the right thing to do here.

base_external_system/readme/CONFIGURE.md
NL66278: No idea. I removed it.

base_external_system/security/ir.model.access.csv
NL66278: No, as it has become an abstract class.

base_external_system/models/external_system.py
NL66278: I minimized the changes now

NL66278:
@thomaspaulb I took up your comments, please look again

base_external_system/tests/test_external_system_adapter.py
```diff
@@ -1,43 +1,24 @@
 # Copyright 2017 LasLabs Inc.
+# Copyright 2023-2024 Therp BV.
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+"""Check wther calling not implemented methods on abstract class raise exceptions."""
```
thomaspaulb: Check whether calling not-implemented methods on abstract classes raises exceptions.

base_external_system/README.rst
```diff
@@ -49,38 +49,55 @@ Configure external systems in Settings => Technical => External Systems
...
-* ``external_destroy_client``: Destroy the connection, if applicable. Does not need
-  to be defined if the connection destroys itself.
+-  ``external_get_client``: Returns a usable client for the system
+-  ``external_destroy_client``: Destroy the connection, if applicable.
+   Does not need to be defined if the connection destroys itself.
+
+Changelog
+=========
+
+Version 16.0.1.0.0
```
thomaspaulb, line 91: Given all the major breaking changes done here, should they also be explained with a `16.0.2.0.0` changelog comment, with a rationale and a migration path? If someone had made custom dependent modules that add fields on `base.external.system`, this will now stop working and someone will need to redo the customization in a new way.

base_external_system/tests/test_external_system_adapter.py
NL66278: Corrected.

base_external_system/README.rst
thomaspaulb: Ok sorry I'm talking nonsense here, it's not `base.external.system` that went to become an AbstractModel. Still, I would appreciate an explanation why this is better, if only to be able to review better, and for users to understand.

NL66278:
@thomaspaulb @ntsirintanis Because of the change (even if remote) that somebody might have extended this module somewhere in a way that is not compatible with these changes, and because this is not fully plugin-compatible, I will close this, but will implement the changes in the migration to 18.0.

NL66278:
The 18.0 migration is here: https://github.com/OCA/server-backend/pull/447

PR #276, opened 2024-03-20
[MIG] base_dav: Migration to 17.0

dreispt:
See https://github.com/OCA/server-backend/pull/269

- [ADD] base_dav
- Fix CI errors
- Mute radicale logger for testing
- Use the collection as first argument in the creation of the items
- Use related model_id field to build domain
- Set default id for fields
- Add missing quotation
- [UPD] Update base_dav.pot
- [UPD] README.rst
- [UPD] README.rst
- [REF] base_dav: pre-commit auto fixes
- [MIG] base_dav: Migration to 17.0

PR #417, opened 2026-02-10
[17.0][ADD] base_dav: port CardDAV module to 17.0

cosmii02:
Based on the base_dav 11.0 module. exports vcard 4.0 and made because of a need for contact book sync to linphone which wants CardDAV address book


StefanRijnhart:
Thanks for your contribution! Unfortunately, the right procedure was not followed and the commit history was lost. Also, there was an earlier attempt that was already quite mature at https://github.com/OCA/server-backend/pull/276, but it was closed due to a lack of interest. I'll reopen it. If you find the time to review that one, or even better, do a functional test, then I think we should be able to get that one merged.

Issue #394, opened 2025-11-27
odoo.sh can't resolve pymssql


creativenetwork-dev:
> breaks possibly a lot of odoo.sh environments 

 _Originally posted by @creativenetwork-dev in [986547b](https://github.com/OCA/server-backend/commit/986547b14de33278f0d42a043394022316854d42#r171509418)_

CRogos:
I agree

```
2025-11-28T20:43:55,797 Created temporary directory: /tmp/pip-unpack-1mizrxmt
2025-11-28T20:43:55,797 Building wheels for collected packages: pymssql
2025-11-28T20:43:55,798   Created temporary directory: /tmp/pip-wheel-w7ibvrc5
2025-11-28T20:43:55,798   Destination directory: /tmp/pip-wheel-w7ibvrc5
2025-11-28T20:43:55,798   Building wheel for pymssql (pyproject.toml): started
2025-11-28T20:43:55,798   Running command Building wheel for pymssql (pyproject.toml)
2025-11-28T20:43:56,213   <string>:32: SetuptoolsDeprecationWarning: The test command is disabled and references to it are deprecated.
2025-11-28T20:43:56,213   !!

2025-11-28T20:43:56,213           ********************************************************************************
2025-11-28T20:43:56,213           Please remove any references to `setuptools.command.test` in all supported versions of the affected package.

2025-11-28T20:43:56,213           This deprecation is overdue, please update your project and remove deprecated
2025-11-28T20:43:56,213           calls to avoid build errors in the future.
2025-11-28T20:43:56,213           ********************************************************************************

2025-11-28T20:43:56,213   !!
2025-11-28T20:43:56,300   /tmp/pip-build-env-2oikye7p/overlay/local/lib/python3.12/dist-packages/setuptools/_distutils/dist.py:289: UserWarning: Unknown distribution option: 'tests_require'
2025-11-28T20:43:56,300     warnings.warn(msg)
2025-11-28T20:43:56,309   /tmp/pip-build-env-2oikye7p/overlay/local/lib/python3.12/dist-packages/setuptools_scm/_integration/version_inference.py:51: UserWarning: version of pymssql already set
2025-11-28T20:43:56,310     warnings.warn(self.message)
2025-11-28T20:43:56,311   /tmp/pip-build-env-2oikye7p/overlay/local/lib/python3.12/dist-packages/setuptools/dist.py:759: SetuptoolsDeprecationWarning: License classifiers are deprecated.
[...]

PR #420, opened 2026-02-11
[18.0][DEP] base_external_dbsource_mssql: proper lib version for trixie

flotho:


CRogos:
@flotho this ticket is also still unsolved: https://github.com/OCA/server-backend/issues/394
Can you check if this module is also working with `pymssql<=2.2.10 ; python_version > '3.10' and python_version < '3.13'`, because then it would be compatible with odoo.sh.

StefanRijnhart (changes requested):
As @CRogos mentions, https://pypi.org/project/pymssql/2.2.8/ is not compatible with Python 3.12 and thus Odoo.sh.

StefanRijnhart:
Thanks! Can you squash commits?

StefanRijnhart:
Obsoleted by https://github.com/OCA/server-backend/pull/437

PR #426, opened 2026-02-27
[18.0][IMP] base_import_match: add Match column to import preview

yostashiro:
Allow users to select match fields directly in the import preview UI.
Checked fields are used to find existing records but are not written.
Pre-configured import match rules pre-check the defaults.

UI selections prevail over configured rules; empty selection skips
matching entirely. When a match finds zero or multiple records, the
entire import is blocked with per-row error messages.

This will allow ad-hoc selection of matching keys at data import.
image

@qrtl QT6195


Sanami0329 (commented):
I have confirmed that there are no issues with functionality.
However, is it expected behavior that the position of Odoo Field shifts (File Column width increases) depending on whether the ID is set as an external ID or not?

Image

Image

Sanami0329 (approved):
LGTM

AungKoKoLin1997 (approved):
Code Review: LGTM

baf-adhoc (approved):
Functional review ok!

PR #428, opened 2026-03-04
[19.0][FIX] base_user_role: Cannot see User Roles in list view, even though the user has roles in form view

namtnt:
**Summary**
Currently, native Odoo already defines a `role_ids` field (m2m). This overrides the `role_ids` field (o2m) from the `base_user_role` module, which causes the User Roles field to be empty in the User list view.

**Changes**
Renamed `role_ids` field to `user_role_ids`.

**Reproduce:**
Create a User and set a role for the User
image

Back to User list view -> User Roles is empty
image

### DIFF
--- a/base_user_role/models/user.py
+++ b/base_user_role/models/user.py
@@ -21,10 +21,10 @@ class ResUsers(models.Model):
         for user in self:
             user.show_alert = user.role_line_ids.filtered(lambda rec: rec.is_enabled)
 
-    role_ids = fields.One2many(
+    user_role_ids = fields.One2many(
         comodel_name="res.users.role",
-        string="User Roles",
-        compute="_compute_role_ids",
+        string="Roles",
+        compute="_compute_user_role_ids",
         compute_sudo=True,
         groups="base.group_erp_manager",
     )
@@ -42,9 +42,9 @@ class ResUsers(models.Model):
         return [{"role_id": r.id} for r in default_roles]
 
     @api.depends("role_line_ids.role_id")
-    def _compute_role_ids(self):
+    def _compute_user_role_ids(self):
         for user in self:
-            user.role_ids = user.role_line_ids.mapped("role_id")
+            user.user_role_ids = user.role_line_ids.mapped("role_id")
 
     @api.model_create_multi
     def create(self, vals_list):
--- a/base_user_role/tests/test_user_role.py
+++ b/base_user_role/tests/test_user_role.py
@@ -208,7 +208,7 @@ class TestUserRole(TransactionCase):
             {"name": "USER TEST (DEFAULT ROLES)", "login": "user_test_default_roles"}
         )
         roles = self.role_model.browse([self.role1_id.id, self.role2_id.id])
-        self.assertEqual(user.role_ids, roles)
+        self.assertEqual(user.user_role_ids, roles)
 
     def test_role_multicompany(self):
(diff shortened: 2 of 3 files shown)


cuongnmtm:
@namtnt  could you please squash the commits? And the FIX/IMP commit msg should not has version prefix, so that we can pw-port down-port it w/o the need to update commit msg.

base_user_role/models/user.py
```diff
@@ -21,9 +21,9 @@ def _compute_show_alert(self):
         for user in self:
             user.show_alert = user.role_line_ids.filtered(lambda rec: rec.is_enabled)
 
-    role_ids = fields.One2many(
+    user_role_ids = fields.One2many(
         comodel_name="res.users.role",
-        string="User Roles",
+        string="Roles",
```
cuongnmtm, line 26: I don't think we need to change this. Do you have a good reason for this?

cuongnmtm (changes requested):

base_user_role/models/user.py
namtnt: We need this change to fix this warning `WARNING odoo odoo.addons.base.models.ir_model: Two fields (user_role_ids, role_ids) of res.users() have the same label: User Roles. [Modules: base_user_role and mail]`

cuongnmtm (approved):

amh-mw (approved):
Code review only. LGTM.

base_user_role/models/user.py
```diff
@@ -21,9 +21,9 @@ def _compute_show_alert(self):
         for user in self:
             user.show_alert = user.role_line_ids.filtered(lambda rec: rec.is_enabled)
 
-    role_ids = fields.One2many(
+    user_role_ids = fields.One2many(
```
dreispt, line 24: Is a migration script needed?

cuongnmtm: No, because the field is a computed non-store field

base_user_role/models/user.py
```diff
@@ -21,9 +21,9 @@ def _compute_show_alert(self):
...
         comodel_name="res.users.role",
-        string="User Roles",
+        string="Roles",
         compute="_compute_role_ids",
```
hoattk-komit: Should we rename the compute method to match the field name? Otherwise, code reviewed and LGTM.

cuongnmtm: Nice suggestion, could you please help do that?

hoattk-komit: done

amh-mw:
Nit: Prefix [19.0] to PR title.

cuongnmtm:
> Nit: Prefix [19.0] to PR title.

We can no longer access the owner's account. Is it possible for the repository maintainer to change the PR title? If not, we could try submitting another PR.

huan-tq (approved):
Funtional Tested and it's LGTM!

CRogos:
@sebalix can you merge?

much-filipe-colaco:
What is the status of this change? Seems too important to leave approved and unmerged for so long.

sebalix (approved):
LG

sebalix:
Thank you for the fix

/ocabot merge minor

sebalix:
I didn't saw the Odoo version number was in the commit message, too late!
As commits are forward ported and backported on others Odoo versions, having it in the commit message is not useful.

Issue #133, opened 2021-09-09
[11.0] base_dav incompatible with Radicale 3.0.0 and up


chrisb-c01:
It seems the Python package Radicale which is used by `base_dav` version 3.0.0 introduces breaking changes making it incompatible with the `base_dav`. Would be good to indicate this/pin Radicale version or make `base_dav` compatible with version 3.0.0. Also see Radicale changelog here: https://github.com/Kozea/Radicale/blob/3.0.0/NEWS.md

Errors I was getting:

```
  File "/../server-backend/base_dav/radicale/auth.py", line 12, in <module>
    class Auth(BaseAuth):
TypeError: NoneType takes no arguments
```

and

```
File "/../server-backend/base_dav/radicale/collection.py", line 26, in <module>
    class FileItem(Item):
TypeError: NoneType takes no arguments
```

antoine45170:
Hello Chris, 
I'm facing the same issue with Radicale 2.1.12. When I tried to launch Odoo server i can the type error for the "class Auth(BaseAuth)". How did you solve the problem ? 
Thank you in advance

chrisb-c01:
@antoine45170 we have pinned radicale to version 2.1.12 in the requirements.txt

PR #431, opened 2026-03-06
[18.0][MIG] base_dav: standard migrate

tendil:
Standard migration from 11.0


cvinh:
Hello,
Thanks for the veeeery welcomed migration !!!
I will review asap.
Do you think that we have any risk to send hundreds of meeting invitations at first sync ? (I remember bad memories about upgrading calendar...)
Best regards

base_dav/__manifest__.py
```diff
@@ -0,0 +1,25 @@
...
+{
+    "name": "Caldav and Carddav support",
+    "version": "18.0.1.0.0",
+    "author": "initOS GmbH,Therp BV,Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "category": "Extra Tools",
+    "summary": "Access Odoo data as calendar or address book",
+    "website": "https://github.com/OCA/server-backend",
+    "depends": [
+        "base",
```
cvinh, line 14: Maybe it should depend on calendar as well... WDYT ?
If yes, then I think it's better to do a separate commit

base_dav/models/dav_collection.py
```diff
@@ -0,0 +1,518 @@
...
+            vobj = result
+        else:
+            return None
+
+        for mapping in self.field_mapping_ids:
+            value = mapping.to_vobject(record)
+            if value is None or value is False:
+                continue
+            if isinstance(value, bool):
+                continue
```
cvinh, line 276:
```suggestion
                continue
             if isinstance(value, list):
                for v in value:
                    if isinstance(v, (int, float)):
                        v = str(v)
                    vobj.add(mapping.name).value = v
                continue
```

cvinh: This allows to manage a list (for example a list of attendees in a calendar.event)
mapping field export_code would be 
`result = ["mailto:" + p.email for p in record.partner_ids if p.email]`

base_dav/readme/CONFIGURE.md
```diff
@@ -0,0 +1,8 @@
+To configure this module, you need to:
+
+1.  Go to Settings / WebDAV Collections and create or edit your
+    collections. There, you'll also see the URL to point your clients
+    to.
+
```
cvinh, line 6: It would be cool to :
- add examples for mapping fields
- recommend using an API Key instead of the user password in the client
- add domain example (for example for personal calendar)

cvinh (changes requested):

base_dav/models/dav_collection.py
```diff
@@ -0,0 +1,518 @@
...
+        return list(safe_eval(self.domain or "[]", self._eval_context()))
+
+    def eval_domain_records(self):
+        """Search records matching the evaluated domain.
+
+        :return: Recordset of matching records
+        :rtype: odoo.models.BaseModel
+        """
+        self.ensure_one()
+        return self.env[self.model_id.model].search(self._eval_domain())
```
cvinh: I have access rights problems when trying to call self.model_id (cannot read ir_model)...
Is the solution to use sudo() ?
```suggestion
        return self.env[self.sudo().model_id.model].search(self._eval_domain())
```
It happens also with self.field_uuid and self.field_id

base_dav/models/dav_collection.py
cvinh: I'm doing a commit to improve that because we need support for list and dicts as well in the calendar case...

base_dav/__manifest__.py
cvinh: Finally, no... improvements for calendar will be in calendar_dav

base_dav/readme/ROADMAP.md
```diff
@@ -0,0 +1,19 @@
...
+  vobject fields);
+- Support todo lists and journals;
+- Support configuring default field mappings per model;
+- Support plain WebDAV collections to make some model's records
+  accessible as folders, and the records' attachments as files (r/w);
+- Support configuring lists of calendars so that you can have a calendar
+  for every project and appointments are tasks, or a calendar for every
+  sales team and appointments are sale orders. Lots of possibilities.
+
+Backporting this to \<=v10 will be tricky because radicale only supports
```
hbrunn: I don't think we need this any more

base_dav/radicale/rights.py
```diff
@@ -1,31 +1,85 @@
...
+
+        :param user: Authenticated username or None
+        :type user: str | None
+        :param path: Requested DAV path
+        :type path: str | None
+
+        :return: Access mode:
+                 - "rw" for read/write
+                 - "r" for read-only
+                 - "" for no access
```
hbrunn: please document values RW

base_dav/radicale/collection.py
```diff
@@ -1,132 +1,438 @@
...
-    def get(self, href):
-        return self.collection.dav_get(self, href)
+        if key == "tag":
+            return self._record.tag
+        if key == "D:displayname":
+            return self._record.display_name or self._record.name
+        if key == "C:supported-calendar-component-set":
+            return "VTODO,VEVENT,VJOURNAL"
+        if key == "ICAL:calendar-color":
+            return "#48c9f4"
```
hbrunn: the TODO here is still valid I think

base_dav/readme/CREDITS.rst
```diff
@@ -1,2 +0,0 @@
-* Odoo Community Association: `Icon <https://github.com/OCA/maintainer-tools/blob/master/template/module/static/description/icon.svg>`_
-* All the actual work is done by `Radicale <https://radicale.org>`_
```
hbrunn: I don't think this should go away

hbrunn (changes requested):
thank you, this looks very good on first sight.

In v18, you can reasonably test this against the caldav lib in a HttpCase, so please do so

ikus060:
Hi @tendil

Thanks a lot for this PR. It comes at a very good timing for me, but it lacks some features I'm counting on. So base on your work, I started filling the gap:

1. Add support for built-in mapper for common use-case: address book mapping & calendar mapping.
2. Allow built-in mapping to be customized.
3. Mapping of calendar recursive event.

It might be a good idea to align each other work, as this work is in progress.

I'm also looking at improving some area like `etag` computation and `sync-token` which currently required the collection to be recomputed every time.

Here is my branch if you want to have a look: https://github.com/ikus060/server-backend/tree/base_dav-built-in-mapper

hbrunn:
@ikus060 this module is called *base*_dav, whatever you do with specific odoo models belongs into an addon depending on this one, and probably into another repository.

It will be helpful if you review the migration as is with possibly an eye on extensibility to allow you doing what you want to do

ikus060:
@hbrunn Thanks for your feedback.

> this module is called base_dav, whatever you do with specific odoo models belongs into an addon depending on this one, and probably into another repository.

Yes, I intend to split my changes into multiple modules in the future. But as I develop it's easier for me to place everything into the same module for now.

> It will be helpful if you review the migration as is with possibly an eye on extensibility to allow you doing what you want to do

The current state is good enough for me. I will submit a new PR when I'm ready.

base_dav/readme/CONFIGURE.md
Aldeigja: > It would be cool to :
> 
> * add examples for mapping fields
> * recommend using an API Key instead of the user password in the client
> * add domain example (for example for personal calendar)

@cvinh  Good point, bit I'm not sure that these are the tasks we should do during module migration, 
I suggest we create a new PR with revised documentation when we merge this PR.

cvinh:
> 1. Add support for built-in mapper for common use-case: address book mapping & calendar mapping.
> 2. Allow built-in mapping to be customized.
> 3. Mapping of calendar recursive event.

Hi @ikus060 
You might be interested in https://github.com/OCA/calendar/pull/209
Best regards

base_dav/readme/CONFIGURE.md
cvinh: sounds great

base_dav/__manifest__.py
```diff
@@ -0,0 +1,25 @@
...
+    "demo": [
+        "demo/dav_collection.xml",
+    ],
+    "data": [
+        "views/dav_collection.xml",
+        "security/ir.model.access.csv",
+    ],
+    "external_dependencies": {
+        "python": ["radicale"],
+    },
```
cvinh, line 25: Missing
"installable" : True,
"application": False,
...?

tendil:
> Hello, Thanks for the veeeery welcomed migration !!! I will review asap. Do you think that we have any risk to send hundreds of meeting invitations at first sync ? (I remember bad memories about upgrading calendar...) Best regards

---

Thanks!

about the "hundreds of invitations" risk: the DAV proxy itself doesn’t send emails - it only forwards CalDAV/CardDAV requests to odoo create/write
to be 100% safe on the first sync, I suggest disabling outgoing mail servers

base_dav/__manifest__.py
tendil: i didn't specify them on purpose because these are the default values - https://arc.net/l/quote/rentkjyl

base_dav/readme/CREDITS.rst
tendil: Thanks for the feedback!
i deleted this file because I followed the migration documentation: https://arc.net/l/quote/bxjruewp

if you'd like it restored, just let me know and I'll revert it to its original state

base_dav/models/dav_collection.py
tendil: Thank you!
I've applied it everywhere needed

base_dav/readme/CREDITS.rst
hbrunn: this is about sentences like "migration to version 42 financially supported by acme ltd", but here we refer to the project that does all of the heavy lifting, so you can just keep the file minus the icon because we don't mention this explicitly any more

cvinh:
LGTM please squash commits and I'll approve

tendil:
> LGTM please squash commits and I'll approve

I did it

cvinh (approved):

base_dav/controllers/main.py
```diff
@@ -0,0 +1,135 @@
...
+        """
+        configuration = radicale_config.load()
+        configuration.update(
+            {
+                "auth": {"type": "odoo.addons.base_dav.radicale.auth"},
+                "storage": {"type": "odoo.addons.base_dav.radicale.collection"},
+                "rights": {"type": "odoo.addons.base_dav.radicale.rights"},
+                "web": {"type": "none"},
+                "hook": {"type": "none"},
+            },
```
hbrunn: take this from a function on self to make it simple to override. it would be perfect if those classes would be AbstractModels, but I don't see a simple way to do that

base_dav/controllers/main.py
```diff
@@ -0,0 +1,135 @@
...
+            "odoo",
+        )
+
+        app = Application(configuration)
+
+        # Let's take WSGI environ from werkzeug/odoo
```
hbrunn: i find those llm-smelling comments somewhat jarring

base_dav/controllers/main.py
```diff
@@ -0,0 +1,135 @@
...
+        # Radicale 3.x requires wsgi.errors and wsgi.input
+        environ.setdefault("wsgi.errors", sys.stderr)
+        method = environ.get("REQUEST_METHOD") or request.httprequest.method
+        raw_body = request.httprequest.get_data(cache=False) or b""
+
+        if method == "PROPFIND" and len(raw_body) == 0:
+            raw_body = (
+                b'<?xml version="1.0" encoding="utf-8"?>'
+                b'<D:propfind xmlns:D="DAV:"><D:allprop/></D:propfind>'
+            )
```
hbrunn: why do we need this? shouldn't radicale handle that? if we need it, it needs a comment why

base_dav/controllers/main.py
```diff
@@ -0,0 +1,135 @@
...
+
+        # Force Radicale to read the body we provide
+        environ["wsgi.input"] = io.BytesIO(raw_body)
+        environ["CONTENT_LENGTH"] = str(len(raw_body))
```
hbrunn: why do you need to set this? should exist already

base_dav/controllers/main.py
```diff
@@ -0,0 +1,135 @@
...
+        raw_body = request.httprequest.get_data(cache=False) or b""
+
+        if method == "PROPFIND" and len(raw_body) == 0:
+            raw_body = (
+                b'<?xml version="1.0" encoding="utf-8"?>'
+                b'<D:propfind xmlns:D="DAV:"><D:allprop/></D:propfind>'
+            )
+
+        # Force Radicale to read the body we provide
+        environ["wsgi.input"] = io.BytesIO(raw_body)
```
hbrunn, line 75: why not

```suggestion
        environ["wsgi.input"] = request.httprequest._HTTPRequest__wrapped.stream
```

base_dav/radicale/auth.py
```diff
@@ -1,18 +1,119 @@
...
 from odoo.http import request
 
-try:
-    from radicale.auth import BaseAuth
-except ImportError:
-    BaseAuth = None
+_logger = logging.getLogger(__name__)
+
+
+def load(configuration):
```
hbrunn: where is this being used?

base_dav/radicale/auth.py
```diff
@@ -1,18 +1,119 @@
...
+    if remote:
+        if isinstance(remote, (tuple | list)):
+            remote = remote[0] if remote else None
+        if remote:
+            env["REMOTE_ADDR"] = str(remote)
+
+    return env
+
+
+def _set_request_user(base_env, uid):
```
hbrunn: please avoid one line functions that are being used once and obfuscate  stuff that's very standard Odoo

base_dav/radicale/auth.py
```diff
@@ -1,18 +1,119 @@
...
+        base_env = request.env
+
+        credential = {
+            "login": login,
+            "password": password,
+            "type": "password",
+        }
+
+        try:
+            res = base_env["res.users"]._login(
```
hbrunn: going through login for every request slows down stuff considerably. i think it will be better to require apikeys and just do `base_env['res.users.apikeys']._check_credentials`. then you also don't need the synthesized user agent stuff.

also `base_env` is an odd name for a variable for an Odoo environment, that canonical name is `env`

base_dav/radicale/auth.py
```diff
@@ -1,18 +1,119 @@
...
+            # Return empty result so Radicale responds with 401 instead of 500
+            _logger.info("DAV login failed for %r", login, exc_info=True)
+            return ""
+
+        uid = res.get("uid") if isinstance(res, dict) else res
+        if not uid:
+            return ""
+
+        _set_request_user(base_env, uid)
+        return request.env["res.users"].sudo().browse(uid).login or login
```
hbrunn: the `or login` fallback is nonsense, remove

hbrunn (changes requested):
on second look, this has some coding assistant smell so i have to switch into bullshit detection mode for review, which is quite taxing. did a couple of files, more to come after followup (please make your own BS removal pass first)

base_dav/models/dav_collection.py
```diff
@@ -0,0 +1,528 @@
...
+            if value is None or value is False:
+                continue
+            if isinstance(value, bool):
+                continue
+            if isinstance(value, (int | float)):
+                value = str(value)
+            vobj.add(mapping.name).value = value
+
+        if "uid" not in vobj.contents:
+            vobj.add("uid").value = f"{record._name},{record.id}"
```
ikus060: I don't understand the logic to generate a UID as `<model_name>,<id>`. This is not symmetric with implementation found in `def get_record()` where we simply expect an "id".

So the right implementation here is to use `id` has the `uid` by default.

In addition, if the collection has a field_uuid, The mapper should use that too.

So the right implementation should read:
```
        if "uid" not in vobj.contents:
            vobj.add("uid").value = (
                record[collection.field_uuid.name]
                if collection.field_uuid
                else str(record.id)
            )
```

base_dav/tests/test_collection.py
```diff
@@ -0,0 +1,321 @@
...
+        collection.delete(record_href)
+        self.assertFalse(self.record.exists())
+
+        new_href = record_href + "0"
+        created_item, _ = collection.upload(new_href, exported)
+        self.assertTrue(created_item)
+
+        # created_item.href can differ from requested href (Odoo assigns new id)
+        vobj = created_item.vobject_item
+        uid_value = ""
```
ikus060: All of the following lines is non-sens too me.

`vobj` should be a vcalendar. We can directly get the uid using:

`uid_value = vobj.contents['vevent'][0].uid`

tendil: agreed! i simplified the uid extraction accordingly

base_dav/models/dav_collection.py
tendil: agreed, thank you! 
i changed the default uid generation to use the same identifier as the collection lookup

base_dav/radicale/auth.py
tendil: switched dav auth to _check_credentials()

i also keep validating that the resolved user login matches the basic auth login

base_dav/radicale/auth.py
tendil: agreed, thank you!

base_dav/radicale/auth.py
tendil: thank you! confirmed: these module-level load() helpers are unused with the current radicale plugin loading flow

radicale loads the configured plugin module and instantiates the expected backend class directly, so these functions were dead code and have been removed

base_dav/controllers/main.py
tendil: request.httprequest.stream is not exposed publicly here

rebuilding wsgi.input from get_data() gives radicale a plain readable stream with a matching CONTENT_LENGTH, independent of how odoo wraps the underlying werkzeug request

this is also the variant i verified locally while request.httprequest.stream is not exposed here

base_dav/controllers/main.py
tendil: wsgi.input is still set explicitly because radicale reads request bodies
from that wsgi key while in this odoo controller bridge it is not exposed in a usable form

if there's a better solution I'd appreciate any suggestions!

diggy128:
This is indeed a very welcomed migration.
Unfortunately, I don't have time to review thoroughly.
From a quick functional review, seems to work OK, and vcf is created.
Couldn't check anything else.

btw. The main reason for migration lagging was that Radicale 3 had breaking changes (see #133).
Does this migration move to radicale 3? In this case maybe pin a version in requirements.txt

base_dav/models/dav_collection_field_mapping.py
```diff
@@ -0,0 +1,407 @@
...
+    vobject,
+    {
+        "vCard": {},
+        "iCalendar": {},
+        "vcard": {"Name": {}},
+        "base": {},
+    },
+)
+
+
```
hbrunn:
```suggestion
```

remove all of that, use the versions of safe_eval directly, wrap vobject in the `_get_safe_eval_context` call below

base_dav/controllers/main.py
hbrunn: as soon as this module fully supports files it will be needed because then we absolutely don't want to copy a potentially large upload here. will not press the point for now

base_dav/controllers/main.py
hbrunn: do nothing and keep whatever content length the client has sent

base_dav/controllers/main.py
```diff
@@ -0,0 +1,106 @@
...
+        app = Application(configuration)
+
+        environ = dict(request.httprequest.environ)
+        environ.setdefault("wsgi.errors", sys.stderr)
+
+        raw_body = request.httprequest.get_data(cache=False) or b""
+        environ["wsgi.input"] = io.BytesIO(raw_body)
+        environ["CONTENT_LENGTH"] = str(len(raw_body))
+
+        environ.setdefault("CONTENT_TYPE", "application/xml; charset=utf-8")
```
hbrunn:
```suggestion
```

same here, don't see the point of fixing wrong clients. to whatever degree that's necessary, it's radicale's job

base_dav/controllers/main.py
```diff
@@ -0,0 +1,106 @@
...
+            """
+            status_headers["status"] = status
+            status_headers["headers"] = headers
+
+        result_iter = app(environ, start_response)
+        try:
+            response_body = b"".join(result_iter) if result_iter else b""
+        finally:
+            if hasattr(result_iter, "close"):
+                result_iter.close()
```
hbrunn:
```suggestion
```

remove this and pass `result_iter` to the response directly

base_dav/controllers/main.py
```diff
@@ -0,0 +1,106 @@
...
+
+        headers = status_headers["headers"]
+        if isinstance(headers, dict):
+            headers = list(headers.items())
```
hbrunn: in which cases do we need that? as in: i suspect we can pass whatever start_response got passed directly

base_dav/models/dav_collection.py
```diff
@@ -0,0 +1,545 @@
...
+        URL is constructed using:
+          - system base URL
+          - DAV prefix
+          - current user login
+          - collection ID
+        """
+        base_url = (
+            self.env["ir.config_parameter"].sudo().get_param("web.base.url") or ""
+        ).rstrip("/")
+        login = self.env.user.login or ""
```
hbrunn:
```suggestion
        login = self.env.user.login
```

base_dav/models/dav_collection.py
```diff
@@ -0,0 +1,545 @@
...
+        for rec in self:
+            rec.url = (
+                f"{base_url}{PREFIX}/{login}/{rec.id}"
+                if base_url
+                else f"{PREFIX}/{login}/{rec.id}"
```
hbrunn:
```suggestion
                f"{base_url}{PREFIX}/{login}/{rec.id}"
```

if base_url is empty, then Odoo is seriously misconfigured, and even if that's the case, both branches will yield the same result. that's what i mean with you should do your own bullshit removal pass, it's your responsibility to do this when you use a bullshit generator, not something you are allowed to offload to the reviewer

base_dav/models/dav_collection.py
```diff
@@ -0,0 +1,545 @@
...
+        :raises Exception: If domain evaluation fails
+        """
+        for rec in self:
+            rec._eval_domain()
+
+    @api.model
+    def _eval_context(self):
+        """Return safe evaluation context for domain expressions.
+
+        :return: Dictionary containing allowed evaluation variables
```
hbrunn:
```suggestion
        :return: Dictionary containing available evaluation variables
```

base_dav/models/dav_collection.py
```diff
@@ -0,0 +1,545 @@
...
+
+    def _eval_domain(self):
+        """Evaluate stored domain expression into Odoo domain list.
+
+        :raises ValueError: If domain string is invalid
+        :return: Evaluated domain
+        :rtype: List[Any]
+        """
+        self.ensure_one()
+        return list(safe_eval(self.domain or "[]", self._eval_context()))
```
hbrunn:
```suggestion
        return safe_eval(self.domain or "[]", self._eval_context())
```

base_dav/radicale/collection.py
```diff
@@ -0,0 +1,390 @@
...
+            rec = env["dav.collection"].browse(int(self.path_components[1])).exists()
+            self._record = rec or None
+
+    def _get_metadata(self):
+        if not self._record:
+            return {}
+
+        metadata = {
+            "tag": self._record.tag or "",
+            "D:displayname": self._record.display_name or self._record.name or "",
```
hbrunn:
```suggestion
            "D:displayname": self._record.display_name or "",
```

we know that this exists

base_dav/radicale/collection.py
```diff
@@ -0,0 +1,390 @@
...
+        }
+
+        if self._record.tag == "VCALENDAR":
+            metadata["C:supported-calendar-component-set"] = "VTODO,VEVENT,VJOURNAL"
+            metadata["ICAL:calendar-color"] = "#48c9f4"
```
hbrunn: it's not much effort to actually implement this now, right?

base_dav/radicale/collection.py
```diff
@@ -0,0 +1,390 @@
...
+
+        :param hrefs: Iterable of href strings
+        :type hrefs: Iterable[str]
+
+        :return: Iterator of (href, item)
+        :rtype: Iterator[Tuple[str, Optional[Any]]]
+        """
+        seen: set[str] = set()
+        for href in hrefs:
+            if href in seen:
```
hbrunn: why do we need this? I'd expect to return exactly what was asked here

base_dav/radicale/collection.py
```diff
@@ -0,0 +1,390 @@
...
+
+    :param collection_path: Collection base path
+    :type collection_path: str
+    :param href: Relative or raw href
+    :type href: str
+
+    :return: Absolute href
+    :rtype: str
+    """
+    h = pathutils.strip_path(pathutils.sanitize_path(href or ""))
```
hbrunn: if you define a one-line function for this, use it consistently, also below

base_dav/views/dav_collection.xml
```diff
@@ -0,0 +1,77 @@
...
+        <field name="arch" type="xml">
+            <form>
+                <group>
+                    <field name="name" />
+                    <field name="dav_type" />
+                    <field name="url" widget="url" invisible="not url" />
+                </group>
+                <group string="Access">
+                    <field name="model_id" />
+                    <!-- TODO: use widget="domain" when we can set the model dynamically /-->
```
hbrunn: this is possible in v18

base_dav/__manifest__.py
```diff
@@ -0,0 +1,25 @@
...
+# Copyright 2019-2020 initOS GmbH <https://initos.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+{
+    "name": "Caldav and Carddav support",
+    "version": "18.0.1.0.0",
+    "author": "initOS GmbH,Therp BV,Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "category": "Extra Tools",
+    "summary": "Access Odoo data as calendar or address book",
+    "website": "https://github.com/OCA/server-backend",
```
hbrunn, line 11:
```suggestion
    "website": "https://github.com/OCA/server-backend",
    "maintainers": ["hbrunn"],
```

if you want to take longterm responsibility for this module, add yourself too

hbrunn (changes requested):
not a fan of the tests, I still think you should use python-caldav to actually query the server instead of the patch fest you do here

same for files and webdavclient3. i'm not aware of a carddav client lib for python3, but that's fine.

there should also be a base class for tests that creates collections, mappings, whatever

base_dav/radicale/collection.py
```diff
@@ -0,0 +1,390 @@
...
+        if not self._record:
+            return {}
+
+        metadata = {
+            "tag": self._record.tag or "",
+            "D:displayname": self._record.display_name or self._record.name or "",
+        }
+
+        if self._record.tag == "VCALENDAR":
+            metadata["C:supported-calendar-component-set"] = "VTODO,VEVENT,VJOURNAL"
```
ikus060: I don't think it makes any send to have a collection that return VTODO and VEVENT.

As we would map either calendar event OR tasks.

What is the use case we are trying to covert here ? 

I would recommand to change a bit the implementation so each collection return either VEVENT, VTODO, VJOURNAL or VCARD, etc.

Take note VJOURNAL is not really supported by anything.

tendil: it no longer advertises mixed component support, so the collection is not misleading anymore and now exposes only VEVENT, which matches the current implementation

support other component types would require separate collection semantics and should be handled in a dedicated follow-up task

base_dav/radicale/collection.py
tendil: agreed

base_dav/radicale/collection.py
tendil: i removed VTODO and VJOURNAL so as not to promote something that doesn't exist
as for implementation, I think it would be better to handle this outside the migration PR as a separate task

we can discuss this here if I haven't fully understood the point of the comment

base_dav/controllers/main.py
tendil: agreed
werkzeug already accepts both mappings and header tuples here, so the dict-to-list conversion was unnecessary
i removed it and now pass the headers through unchanged

base_dav/radicale/collection.py
hbrunn: the point of the comment is: don't hardcode a color, but add a field on dav.collection that allows the user to set a color

tendil: oh, right
already done that

base_dav/controllers/main.py
```diff
@@ -0,0 +1,95 @@
...
+        :type davpath: str | None
+        :param kwargs: Extra route keyword arguments, unused.
+        :type kwargs: dict
+
+        :return: Response produced by Radicale.
+        :rtype: odoo.http.Response
+        """
+        configuration = radicale_config.load()
+        configuration.update(self._get_radicale_config(), "odoo")
+        app = Application(configuration)
```
ikus060: @tendil I'm not sure this implementation is optimal. The application get re-created on every request. I think it should be turned into a single. What do you think ?

tendil: agreed 
i changed it to a lazily cached app instance at module level, so it is created once per odoo worker process instead of once per request

i used a module-level cache instead of controller state to keep the lifecycle explicit and make it easy to reset in tests

ikus060: @tendil Thanks for the update. I will test it soon.

I'm poking around because I'm currently working to implement CalDAV for calendar events on top of your work. :smile:

diggy128:
Any update on this?

tendil:
> Any update on this?

---

yes, the pr is ready for review; you can take a look

however, after the rebase, there are issues with the tests in the `base_external_system` module
we are investigating the cause

tendil:
@hbrunn, i would really appreciate it if you could review the code again!

the test pipeline has been fixed

ikus060:
Hi @tendil,
As already mention, I'm working on top of your implementation to provide built-in mapper for contact and calendar event. While most of the mapping is now complete on my side, I started running some test and I'm facing a couple of performance bottleneck mostly cause by the implementation of collection.

Are you interest in fixing them on your side ? To you have time for this ? Or you prefer I submit additional PR ?

---

On every request made, GET http://localhost:8069/.dav/admin/2
The mapping is called twice: (1) by collection.serialize() and (2) another one by collection.etag. The best way I could think of to avoid this dual computation is to implement the etag differently.

```
from hashlib import blake2b

class Collection(BaseCollection):

    @property
    def etag(self) -> str:
        """
        Return etag for the collection. Delegate this to dav.collection.
        """
        if not self._record:
            raise NotImplementedError()
        return self._record.dav_etag()
...

class DavCollection(models.Model):

    def dav_etag(self) -> str:
        """
        Return a ETAG to represent this collection. 
        """
        self.ensure_one()

        # Use blake2b, because it's fast.
        h = blake2b(digest_size=16)

        # Compute hash using id + write_date for all record- including the collection it self.
        h.update(bytes(self.id))
        h.update(str(self.write_date).encode())

        model_name = self.sudo().model_id.model
        rows = self.env[model_name].search_read(
            self._eval_domain(),
            fields=["id", "write_date"],
            order="id asc",  # stable ordering
        )
        for row in rows:
            h.update(bytes(row['id']))
            h.update(str(row['write_date']).encode())

        return '"%s"' % h.hexdigest()
```

---

The other bottleneck I found is the implementation around `get()`, `get_all()`, `get_multi()` which effectively make a SQL query for every single record instead of a single query to list all the record at once.
When you have ~100 of record I guess it's not an issue, but with 1000 record, the performance start to drop quickly, making it unusable for calendar.event.

I think the `get()`, `get_all()` and `get_multi()` should only trigger a single SQL query to fetch everything from the database at once.

tendil:
> Hi @tendil, As already mention, I'm working on top of your implementation to provide built-in mapper for contact and calendar event. While most of the mapping is now complete on my side, I started running some test and I'm facing a couple of performance bottleneck mostly cause by the implementation of collection.
> 
> Are you interest in fixing them on your side ? To you have time for this ? Or you prefer I submit additional PR ?
> 
> On every request made, GET http://localhost:8069/.dav/admin/2 The mapping is called twice: (1) by collection.serialize() and (2) another one by collection.etag. The best way I could think of to avoid this dual computation is to implement the etag differently.
> 
> The other bottleneck I found is the implementation around `get()`, `get_all()`, `get_multi()` which effectively make a SQL query for every single record instead of a single query to list all the record at once. When you have ~100 of record I guess it's not an issue, but with 1000 record, the performance start to drop quickly, making it unusable for calendar.event.
> 
> I think the `get()`, `get_all()` and `get_multi()` should only trigger a single SQL query to fetch everything from the database at once.

---

hello @ikus060 !
thank you for the analysis;
i did run into this during testing, yes, but as part of the migration, I decided not to optimize it right now, since the main goal is to migrate the module correctly;
and after this PR is merged, I plan to tackle the optimization in separate PRs;

i suggest we wait for the author of this module to review it and provide their comments on the matter

if it’s decided to implement the optimization in this PR - I’ll do it;
if it’s confirmed that it needs to be done in separate PRs - I’ll do it in the next PR

ikus060:
Thanks @tendil for your quick reply.

> i suggest we wait for the author of this module to review it and provide their comments on the matter

Are we waiting for @hbrunn to make the review ? or anybody else ?

tendil:
> Thanks @tendil for your quick reply.
> 
> > i suggest we wait for the author of this module to review it and provide their comments on the matter
> 
> Are we waiting for @hbrunn to make the review ? or anybody else ?

---

yes, @hbrunn is the author of this module, and his review and decision will be decisive

ikus060:
No activity since April. Anything I could do to get this PR accepted and merged ?

diggy128:
Second it. 
I've using this module for about 40 days in production with no major issues.
Functional testing seems OK.

hbrunn:
> Are we waiting for @hbrunn to make the review ? or anybody else ?

we are waiting for all of my points being addressed and not every commit adding new bullshit that makes me not want to waste more time reviewing this

ikus060:
> we are waiting for all of my points being addressed and not every commit adding new bullshit that makes me not want to waste more time reviewing this

@hbrunn I understand the frustration, and I want to acknowledge that some of the points you've raised are valid and worth addressing.

That said, I think it's worth stepping back and clarifying the scope of this PR. The goal here is to migrate base_dav to 18.0, not to make it feature-complete or address every possible improvement in one go. If we try to solve everything in a single PR, we risk it never getting merged, which blocks everyone who needs this module to move forward.

My suggestion:

1. Let's agree on what's actually blocking this pull request vs what's an enhancement/improvement that can be handled separately.
2. I already have a follow-up branch in progress to address further improvements, and more are planned after that. This PR is meant to be a solid starting point, not the final state.

If there are specific must-fix items you consider blockers for this PR, could you list them explicitly? That way we can focus the review on those, and track the rest as follow-up issues/PRs so nothing gets lost.

I want to be respectful of your review time and effort — the goal is to converge, not to keep expanding scope indefinitely.

jans23:
I didn't look into detail but I'm using [base_ical](https://github.com/OCA/server-backend/tree/18.0/base_ical) successfully and have the impression that it may be more promising to enhance base_ical instead of trying to bring base_dav further.

ikus060:
@jans23  Thanks for the tips about `base_ical`. I did not know about it. It look promissing, but the simple fact, it doesn't handle CardDav is a blocker for me. I'm already using CardDav and CalDav with base_dav.

That said, it might be beneficial for everyone working on "dav" implementation (webdac, caldav and cardav) to aligned their effort together. I did not look into all the details implementation of `base_ical`, but I know the main culprit to implement caldav is the mapping.

I wasted a good amounth of hours to properly map calendar.event to caldav in this branch. https://github.com/ikus060/calendar/blob/18.0/calendar_dav/models/calendar_event_mapper.py

Handling of rrule is such a pain.

PR #167, opened 2022-05-25
[13.0][FIX]base_user_role: Show active users on Users page

GuillemCForgeFlow:
On the tree view of Roles, the number of users that is being shown is for active users, but once opening a Role, on the Users page of the form view, there are more users as it wasn't filtering for the active ones.

If there's a more elegant way to do it, please don't hesitate to correct me :smile:

PR #444, opened 2026-03-30
[18.0][FIX] base_user_role: remove active field from res.users.role.line

kanda999:
The `active` field was added in PR #167  to exclude archived users from the role list view count. However, it causes Odoo's active_test filtering to silently hide role lines of archived users from the role form view, making it impossible to audit role assignments.

Replace it with a domain on `line_ids` to filter archived users without the active_test side effect.

@qrtl QT6433

### DIFF
--- a/base_user_role/models/role.py
+++ b/base_user_role/models/role.py
@@ -21,7 +21,10 @@ class ResUsersRole(models.Model):
         string="Associated group",
     )
     line_ids = fields.One2many(
-        comodel_name="res.users.role.line", inverse_name="role_id", string="Role lines"
+        comodel_name="res.users.role.line",
+        inverse_name="role_id",
+        string="Role lines",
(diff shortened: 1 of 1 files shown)


base_user_role/models/role.py
```diff
@@ -134,7 +137,6 @@ class ResUsersRoleLine(models.Model):
     _name = "res.users.role.line"
     _description = "Users associated to a role"
 
-    active = fields.Boolean(related="user_id.active")
```
AungKoKoLin1997, line 137: Do we still need this filtering?
https://github.com/OCA/server-backend/blob/ff48dc794098d1c6f441fb0a96a56ba5e302d030/base_user_role/models/user.py#L37

StefanRijnhart (commented):
Would it work to set context active_test=False in the line_ids field view element?

PR #452, opened 2026-04-20
[18.0] Add base_debug_restricted (restrict Debug Mode with a group)

guewen:
By default, Debug Mode is even available to public users on the web part. When
enabled and an error is raised, the public user will see a detailed error page with
the traceback and file paths. It discloses information about the server, which
may be avoided.

This can be seen for example by provoking a 500 by accessing `/web/reset_password?debug=1&self=test` on a runbot database.

image

After installing this module, only connected users with the "Debug Mode" role
will be able to enable the Debug Mode and see detailed error pages.


amh-mw (changes requested):
Doing some functional testing locally, I'm not seeing any difference in behavior between `admin` user and non-admin. Both are getting

> 403 Forbidden
> You can only call user.has_group() with your current user.

for http://localhost/odoo/sales?debug=1.

guewen:
Oh wow, thanks @amh-mw, indeed, not sure why it worked on `/web`. Fixed with a `.sudo()`.

amh-mw (changes requested):
Confirmed absence of 403 Forbidden errors.

Commit title should be of form `[ADD] base_debug_restricted: ...`

For http://localhost/odoo/dashboards?debug=1&dashboard_id=1, lots of console.debug lines being emitted, _i.e._

```
##### Model creation #####
### Loading data ###
Migrating data from version 21
Data migrated in 1 ms
Data loaded in 1 ms
###
Replayed 0 commands in 0 ms
evaluate all cells 11 ms
START 25 ms
Model created in 47 ms
######
evaluate all cells 14 ms
EVALUATE_CELLS 14 ms
evaluate all cells 9 ms
EVALUATE_CELLS 9 ms
```

Though I am not concerned about these messages specifically, it does make me wonder more generally if there is anything concerning making it into the console logs elsewhere. This seems potentially out of scope for this pull request, but related.

guewen:
@amh-mw I amended my commit.

I installed a fresh 18.0 with `-isale_management,website` (so without `base_debug_restricted` and opening `/odoo/dashboards?dashboard_id=1`, even without `debug=1` printed these logs as well.

image

Actually, they match `console.debug()` calls in https://github.com/odoo/odoo/blob/18.0/addons/spreadsheet/static/src/o_spreadsheet/o_spreadsheet.js which are called unconditionally.

amh-mw (approved):

PR #453, opened 2026-04-22
[18.0][FIX] base_user_role: Define users with different names to avoid potential side effects

victoralmau:
Define users with different names to avoid potential side effects

@Tecnativa

### DIFF
--- a/base_user_role/tests/test_user_role.py
+++ b/base_user_role/tests/test_user_role.py
@@ -58,7 +58,7 @@ class TestUserRoleCommon(BaseCommon):
         # Setup for multi-company testing
         cls.multicompany_user_1 = cls.user_model.create(
             {
-                "name": "User 2",
+                "name": "multicompany_user_1",
                 "company_id": cls.company1.id,
                 "company_ids": [(6, 0, [cls.company1.id, cls.company2.id])],
                 "groups_id": [(6, 0, cls.env.ref("base.group_erp_manager").ids)],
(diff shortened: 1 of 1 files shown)


eduezerouali-tecnativa (approved):

amh-mw:
> Define users with different names to avoid potential side effects

What are the potential side effects? It should be allowed that two users have the same name?

victoralmau:
> > Define users with different names to avoid potential side effects
> 
> What are the potential side effects? It should be allowed that two users have the same name?

Exactly, multiple users with the same name; for example, if you use `dms_field` (which automatically creates directories), that error would occur, which doesn't happen under normal circumstances.

amh-mw:
> Exactly, multiple users with the same name; for example, if you use `dms_field` (which automatically creates directories), that error would occur, which doesn't happen under normal circumstances.

So at a large company, `dms_field` doesn't support having two users both named John Smith? It seems pretty dubious to have any part of the file system depend on user-entered names.

victoralmau:
> > Exactly, multiple users with the same name; for example, if you use `dms_field` (which automatically creates directories), that error would occur, which doesn't happen under normal circumstances.
> 
> So at a large company, `dms_field` doesn't support having two users both named John Smith? It seems pretty dubious to have any part of the file system depend on user-entered names.

This behavior occurs only if a template is defined that automatically creates a directory (using the partner's name) when a partner is created (something that doesn't happen in a “normal” environment), and the restriction that multiple directories cannot have the same name is imposed by DMS; In any case, the “problem” is clear—do not create multiple users with the same name—and the solution is also clear, in my opinion.

pedrobaeza (approved):
This is something only for integration tests, so it's not a problem to use other user names there. 

/ocabot merge patch

PR #382, opened 2025-10-09
[19.0][MIG] base_user_role: Migration to 19.0

tishmen:
Scope

- base_user_role only

Summary

- Migrate to Odoo 19.0 following OCA guide.
- API updates: `group_ids` rename, `SUPERUSER_ID` import from `odoo.api`.
- Switch to `models.Constraint` and `models.Index` APIs.
- Switch to `res.groups` `privilege_id` (drop `category_id`).
- Replace `trans_implied_ids` usages with `all_implied_ids` where effective groups are expected.
- Adjust search/tree view xpaths and `<search><group>` schema for 19.0.
- Refactor default roles due to removal of default user in Odoo.

Pre-commit

- Verified locally; auto-fixes applied and committed.

Tests

Command:

```
./odoo/odoo-bin -c odoo.conf -d bur19_tests_bur -i base_user_role --test-enable --test-tags "/base_user_role" --without-demo=all --log-level=info --stop-after-init
```

Result: green locally.

PR #456, opened 2026-05-10
[FIX] base_user_role: don't reset notification_type

mprechl:
## Problem

The 19.0 migration of `base_user_role` dropped the
`_get_self_writable_groups` method that was present in 18.0. This method
excluded `mail.group_mail_notification_type_inbox` from the role
enforcement logic in `set_groups_from_roles`.

The result is that any user with roles assigned can never persistently set
their notification preference to "In Odoo" — it gets reset on every
`write()` call to `res.users`, or at latest on the next run of the
`cron_update_users` scheduled action.

## Solution

Restore `_get_self_writable_groups` from the 18.0 implementation and use
it in `set_groups_from_roles` to exclude `mail.group_mail_notification_type_inbox`
from both sides of the group diff, exactly as was done before the
migration.

## References

- Migration PR that introduced the regression: #382

### DIFF
--- a/base_user_role/__manifest__.py
+++ b/base_user_role/__manifest__.py
@@ -4,7 +4,7 @@
 
 {
     "name": "User roles",
-    "version": "19.0.1.0.2",
+    "version": "19.0.1.0.1",
     "category": "Tools",
     "author": "ABF OSIELL, Odoo Community Association (OCA)",
     "license": "LGPL-3",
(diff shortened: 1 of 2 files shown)


sebalix:
Hello @mprechl , thank you for the bug fixing, but this is not the migration to 19.0 which dropped the fix, the module was simply fixed in 18.0 after it has been migrated to 19.0:

- https://github.com/OCA/server-backend/pull/372

In such case, the commit(s) of this PR should be cherry-picked on 19.0 properly.

sebalix:
I open #479 to port several PRs missing in 19.0. Closing this one, thanks :pray:

PR #457, opened 2026-05-21
[ADD] New module base_import_strip

legalsylvain:
Rational : Remove spaces at the beginning and end of all values, when importing. So if a spreadsheet contains the value ' France ' as a country_id, the country will be found


remi-filament (approved):
Runboat is down, but code review LGTM, thanks @legalsylvain that will make the import faster !

fcayre (approved):
Code review OK. Thank you !

PR #462, opened 2026-06-11
[18.0][FIX] base_user_role_profile: use self.env in session_info to support nodb authenticate

RicardCForgeFlow:
```
odoo-1                                         |   File "/opt/odoo/auto/addons/base_user_role_profile/models/ir_http.py", line 11, in session_info
odoo-1                                         |     user = request.env.user
odoo-1                                         |            ^^^^^^^^^^^^^^^^
odoo-1                                         | AttributeError: 'NoneType' object has no attribute 'user'
```

occurs during stateless JSON-RPC authentication calls (e.g., from external automated scripts or API clients hitting `/web/session/authenticate`).

### DIFF
--- a/base_user_role_profile/models/ir_http.py
+++ b/base_user_role_profile/models/ir_http.py
@@ -8,7 +8,7 @@ class Http(models.AbstractModel):
 
     def session_info(self):  # pragma: no cover
         result = super().session_info()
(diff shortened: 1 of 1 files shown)


StefanRijnhart (commented):
Would it be possible to add a test?

PR #469, opened 2026-07-02
[19.0][FIX] base_user_role: make role_user_ids searchable

AaronHForgeFlow:
Searching by user in the res.users.role list/search view raised 'Cannot convert res.users.role.role_user_ids to SQL because it is not stored' because role_user_ids is a non-stored computed One2many.

Add a search method redirecting the query to the stored line_ids.user_id relation.


MiquelRForgeFlow (approved):

LoisRForgeFlow (approved):
LGTM

BhaveshHeliconia (approved):

StefanRijnhart (approved):

sebalix (approved):

PR #476, opened 2026-08-11
[18.0][IMP] base_user_role: add a sequence on role and role line

nayatec:
When there is a few roles, created upon time, it may becomes quite handy to be able to sort them is a "logical" order like with increased rights or so.
This PR allows it.

### DIFF
--- a/base_user_role/models/role.py
+++ b/base_user_role/models/role.py
@@ -12,7 +12,9 @@ class ResUsersRole(models.Model):
     _name = "res.users.role"
     _inherits = {"res.groups": "group_id"}
     _description = "User Role"
+    _order = "sequence,id"
 
+    sequence = fields.Integer(default=10)
     group_id = fields.Many2one(
         comodel_name="res.groups",
         required=True,
@@ -133,7 +135,9 @@ class ResUsersRole(models.Model):
 class ResUsersRoleLine(models.Model):
     _name = "res.users.role.line"
     _description = "Users associated to a role"
+    _order = "sequence,id"
 
+    sequence = fields.Integer(related="role_id.sequence")
     active = fields.Boolean(related="user_id.active")
     role_id = fields.Many2one(
         comodel_name="res.users.role", required=True, string="Role", ondelete="cascade"
--- a/base_user_role/models/role.py
+++ b/base_user_role/models/role.py
@@ -12,7 +12,7 @@ class ResUsersRole(models.Model):
     _name = "res.users.role"
     _inherits = {"res.groups": "group_id"}
     _description = "User Role"
-    _order = "sequence,id"
+    _order = "sequence,name"
 
     sequence = fields.Integer(default=10)
     group_id = fields.Many2one(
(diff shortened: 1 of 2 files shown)


base_user_role/models/role.py
```diff
@@ -133,7 +135,9 @@ def show_model_access_ids(self):
 class ResUsersRoleLine(models.Model):
     _name = "res.users.role.line"
     _description = "Users associated to a role"
+    _order = "sequence,id"
 
+    sequence = fields.Integer(related="role_id.sequence")
```
amh-mw, line 140: Does this actually work for a non-stored sequence? Doesn't `_order` govern SQL generation?

amh-mw (changes requested):
Code review only, LGTM.

base_user_role/models/role.py
nayatec: Well, i admit i didn't dug too much into the ORM, but this is the way i found to have role lines sorted according to the roles's sequence, and it actually work.

amh-mw (approved):

base_user_role/models/role.py
```diff
@@ -12,7 +12,9 @@ class ResUsersRole(models.Model):
     _name = "res.users.role"
     _inherits = {"res.groups": "group_id"}
     _description = "User Role"
+    _order = "sequence,id"
```
legalsylvain: could you put sequence,name instead. so if user doesn't want to use sequence to order, the default alphabetical order is used.

thanks !

legalsylvain (changes requested):
a remark inline. otherwise, LGTM.

base_user_role/models/role.py
sebalix: Agree

nayatec: Yes very nice idea !
Done.

sebalix (approved):

base_user_role/models/role.py
```diff
@@ -135,9 +135,12 @@ def show_model_access_ids(self):
 class ResUsersRoleLine(models.Model):
     _name = "res.users.role.line"
     _description = "Users associated to a role"
-    _order = "sequence,id"
+    _order = "sequence,name"
 
     sequence = fields.Integer(related="role_id.sequence")
+    # name is used in _order (see above) to sort lines
+    # the same way roles are sorted.
+    name = fields.Char(related="role_id.name")
```
legalsylvain, line 143: if you don't store it, it is not possible to order by. don't you think ?

base_user_role/views/role.xml
```diff
@@ -75,7 +75,7 @@
         <field name="name">res.users.role.list</field>
         <field name="model">res.users.role</field>
         <field name="arch" type="xml">
-            <list>
+            <list default_order="sequence,name">
```
legalsylvain, line 78: why reorder by the same order defined in python ?

legalsylvain (changes requested):

base_user_role/models/role.py
nayatec: Well, as well as with sequence, it actually works...

base_user_role/views/role.xml
nayatec: Because i saw that when we use the handle widget the name is not taken into account for sorting. So this is the trick i found. When sequence is simply shown without the handle widget, this is not needed however (but then useless from the user's perspective).

If there is a better way, i would happily take it !

base_user_role/models/role.py
legalsylvain: OK !

base_user_role/views/role.xml
sebalix: Maybe add a comment on top explaining the intent of this change as it's not obvious, in a future version the Odoo behavior could change I suppose so we could remove it.

legalsylvain: ok !

PR #477, opened 2026-08-11
[18.0][IMP] base_user_role_profile: add a sequence on role profile

nayatec:
When there is a few profiles, created upon time, it may becomes quite handy to be able to sort them is a "logical" order like with increased rights or so.
This PR allows it.

### DIFF
--- a/base_user_role_profile/models/res_users_profile.py
+++ b/base_user_role_profile/models/res_users_profile.py
@@ -6,7 +6,9 @@ from odoo import fields, models
 class ResUsersProfile(models.Model):
     _name = "res.users.profile"
     _description = "Role profile"
+    _order = "sequence,id"
 
+    sequence = fields.Integer(default=10)
     name = fields.Char()
     user_ids = fields.Many2many(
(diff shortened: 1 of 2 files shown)


amh-mw (approved):
Code review only, LGTM.

base_user_role_profile/models/res_users_profile.py
```diff
@@ -6,7 +6,9 @@
 class ResUsersProfile(models.Model):
     _name = "res.users.profile"
     _description = "Role profile"
+    _order = "sequence,id"
```
sebalix: Same review than https://github.com/OCA/server-backend/pull/476 applies here I guess :)

nayatec: Yes i thought so too, but after trying it this morning i was not convinced of the added value of it.
I can commit the modifs though. I'll do that.

nayatec: Done.

sebalix (approved):

Issue #279, opened 2024-03-28
base_user_role_company doesn't work with multiple browser tabs


bguillot:
Hello, 

On module base_urser_role_company

If several tabs are open on different companies, the role is not correctly assigned.

1. Create a database with Company A and Company B,
2. Create 2 roles : Role A  and Role B
3. Assign a user Role A to company A and Role B to company B
4. open a first tab, select Company A the user has the Role A
5. open a second tab, select company B
6. Go back to first tab, the user has role B

Having looked at how the base_user_role module works, I can't really see a solution to fix this other than disabling multi-tabbing.
Perhaps we should put in the known issues that this problem exists unless someone has a solution!

Thanks for your comments on this issue.

PR #478, opened 2026-09-09
[19.0][FIX] base_user_role_company: fix cross-company privilege escalation …

arnaudlayec:
…and migrate 'role_line.company_id' to 'company_ids'

**Issue description**
Formerly the `res.users.group_ids` was recomputed and written in DB when a user was browsing from a company to another. This implementation was letting a security risk:
- a user could let a former browser tab opened with the previous company opened: when browsing this former tab, it was with former company BUT with the rights of the NEWLY opened company in the NEWER tab
- by API, it was possible to authenticate a user on a Company A and then browse a Company B with the permissions of Company A

**Fix**
Opportunity:
Since v19.0, a new **computed** field `all_group_ids` exist on `res.groups`. This field is called by `ir.model.access::_get_allowed_models` and ir.rule::_compute_domain`.

Solution:
The fix inherits `_compute_all_group_ids` to make it compatible with company-aware roles. It also adapts calls to @tool.ormcache, so the proper caches are used / unvalidated when `self.env.companies` changes.
Note: there is nothing to do on `ir.rule` because the cache is already company-dependant.

**Necessary changes**
- Removal of `ir.http::session_info`, which was writing the `res.users:group_ids`: this is now obsolete, and only done at `res.users::create/write` for UI purpose (and only base on user's main `company_id`)
- Controller `web_load_menus`: add the missing context key `allowed_company_ids`, needed to compute the proper environment var `self.env.companies`. If not done, the navbar menu is generated based on the user's default `company_id` groups (`all_group_ids`). In other word: the menu item were not dynamically changing when the user was browsing other companies.

**Side changes / other improvements**
- Simplification of the code for INTERSECTION logic, when 2 or more companies are active
- File renaming, to follow conventions
- I also migrated `role_line.company_id` to a `company_ids` fields, more convenient


arnaudlayec:
Note: this is related to this former issue https://github.com/OCA/server-backend/issues/279
Ping @bguillot @florian-dacosta @dreispt

dreispt:
Thank you for the work.
Refactoring/reorganizing code at the same time sensitive changes are introduced is not helpful.
Please either do those in separate commits, or simply do not reorganize code and make changes directly in the original files.

Also, you mention in a separate thread that the multicompany group INTERSECT logic was implemented.
Then why isn't it working?
Could this simply be a cache invalidation that needs to happen when switching between companies?

arnaudlayec:
> Thank you for the work. Refactoring/reorganizing code at the same time sensitive changes are introduced is not helpful. Please either do those in separate commits, or simply do not reorganize code and make changes directly in the original files.
> 
> Also, you mention in a separate thread that the multicompany group INTERSECT logic was implemented. Then why isn't it working? Could this simply be a cache invalidation that needs to happen when switching between companies?

Thanks.
OK, I will split in several commits.
The security fix is touching almost everything in the module though, so the split is not that obvious everytime.

Regarding the INTERSECT, I do not fully understand your question.
It is indeed already [implemented here](https://github.com/OCA/server-backend/blob/18.0/base_user_role_company/models/user.py#L39). It is working fine, this logic should not be changed in my opinion. It is a very separated topic that the security issue raised here.
In my commit in this PR, I just refactorized this code ([here](https://github.com/OCA/server-backend/pull/478/changes#diff-2d7182c1671d09cd0977ab0c16550c4ec86afa4bd177e1a4e39b24aaaa760672R45)) because I think it could be more readable. This is a thing I will isolate in a separate commit later.

arnaudlayec:
Hello @dreispt , I splitted the changes in 4 commits, to distinguish the fix from the other changes
This is indeed more visible

base_user_role_company/models/user.py
```diff
@@ -18,23 +18,28 @@ def authenticate(self, *args, **kwargs):
...
-                        active_roles |= role_line
-            return active_roles
-        return res
+        if not self.role_line_ids:
+            return res
+
+        if self.env.context.get("active_company_ids"):
+            company_ids = set(self.env.context.get("active_company_ids"))
+        else:
+            company_ids = set(self.company_id.ids)
```
dreispt, line 33: This can be one line:
```
        company_ids = self.env.context.get("active_company_ids") or self.company_id.id
```

arnaudlayec: This is in the 1st commit but it's actually replaced in the 2nd commit, see [here](https://github.com/OCA/server-backend/pull/478/changes/49b7d059406743f1f9681f8042d206a296686d0d#diff-fba01acaea5745c663526280c64a9ea96d3d77c9dfdbaeb536c26be2ea88160aR44).

`self.env.context.get("active_company_ids")`
This is replaced by `self.env.companies.ids`. Shorter, cleaner. Used in the new method `_get_company_aware_roles`, called to compute `all_group_ids`, used in access checks.

`self.company_id`
This is removed from any logics related to access control checks: we should never default user's access to its default company.
This is now only used in the former existing method `_get_enabled_roles`, to write in the `group_ids` of a user, and display _something (some groups of 1 of its companies)_ in user's form. So this is for layout only.

arnaudlayec:
Hello
May I kindly follow-up about this review
Thanks!

PR #480, opened 2026-09-14
[18.0][ADD] record_trace: get unlinked records trace

bealdav:


amh-mw (changes requested):
Redundant with https://github.com/OCA/server-tools `auditlog` module?

